PR #10, opened 2019-05-15
Add /install route for installing watch-keeper

adamkingit:


app/routes/install/index.js
```diff
@@ -35,47 +35,48 @@ router.use(asyncHandler(async (req, res, next) => {
...
-  org.isUpdate = isUpdate;
-  org.uuid = uuid();
 
-  const template = org.orgYaml;
+router.get('/inventory', asyncHandler(async(req, res, next) => {
+  const apiKey = req.apiKey;
+  var razeeapiUrl = `${req.protocol}://${req.get('host')}/api/v2`;
+  const wk_url = 'https://github.com/razee-io/watch-keeper/releases/latest/download/resource.yaml';
+  try {
+    const  inventoryBuffer = await readFile(`${__dirname}/inventory.yaml`);
```
alewitt2:
```suggestion
    const  inventoryBuffer = await readFile(`${__dirname}/inventory.yaml`, 'utf8');
```
should be able to specify encoding to just get a string instead of the buffer. [nodejs docs] (https://nodejs.org/api/fs.html#fs_fs_readfile_path_options_callback)

alewitt2: if your npm package doesnt support the encoding option, i know `fs-extra` does

PR #12, opened 2019-05-16
Add kapitan install routs

alewitt2:


app/routes/install/index.js
```diff
@@ -75,5 +99,17 @@ router.get('/inventory', asyncHandler(async(req, res, next) => {
   }
 }));
 
+router.get('/kapitan/:component', asyncHandler(async (req, res, next) => {
+  const kptn_url = `https://github.com/razee-io/${req.params.component}/releases/latest/download/resource.yaml`;
+  try {
+    const kptn = await request.get(kptn_url);
+    res.setHeader('content-type', 'application/yaml');
+    return res.status(200).send(kptn);
```
adamkingit: Probably need to pass the status from github through instead of wiring it to 200

alewitt2: done

PR #14, opened 2019-05-16
Create indexes merge

adamkingit:


adamkingit:
Updated code 
- ensures collections and indexes are created once for each time the server starts
- Does not start listening until all the db init code is complete
- Gets the database connection once and sets it on the app
- Accepts orgKey as either a request header or query param for routes that require it

app/routes/index.js
```diff
@@ -16,30 +16,81 @@
...
+router.use('/api/kube', Kube);
+router.use(ebl(getBunyanConfig('/api/v2/')));
+
+router.use(asyncHandler(async (req, res, next) => {
+  const db = req.app.get('db');
+  req.db = db;
+  next();
+}));
 
 router.use((req, res, next) => {
```
dalehille: are we intending to make `razee-org-key` required for the the `/api/kube` route ?

adamkingit: No. /api/kube will be the only route that does not require an orgKey

PR #16, opened 2019-05-20
Change orgs document from apiKey:String to orgKeys:[String]

adamkingit:


adamkingit:
Run the following against your local mongo before running this code:
``` 
db.getCollection('orgs').find({}).forEach((org)=>{
    print(org.apiKey)
    db.orgs.updateOne({ _id: org._id }, { $set: { orgKeys: [org.apiKey] } });
})
```

PR #55, opened 2019-07-16
Use S3 to store resource data if s3 config endpoint is defined.

geoffcorey:


app/routes/v2/clusters.js
```diff
@@ -52,6 +53,14 @@ const addUpdateCluster = async (req, res, next) => {
     next(err);
   }
 };
+const pushToS3 = async (req, key, dataStr) => {
+  //if its a new or changed resource, write the data out to an S3 object
+  const bucket = `razee_${key.org_id}`;
+  const hash = crypto.createHash('sha256');
+  const hashKey = hash.update(JSON.stringify(key)).digest('hex');
+  await req.s3.createBucketAndObject(bucket, hashKey, dataStr);
+  dataStr = `https://${req.s3.endpoint}/${bucket}/${hashKey}`;
```
dalehille: needs a `return` statement

geoffcorey: done

PR #69, opened 2019-08-02
Trust all proxies

mckaymic:


mckaymic:
Apparently IKS Ingress doesn't use any of the default proxy subnets https://expressjs.com/en/guide/behind-proxies.html

PR #78, opened 2019-08-21
add swagger

### LINKED ISSUE #77: Missing API documentation
The API service should have an advertised API spec so other's can integrate with the system. The service should use Swagger/OpenAPI https://swagger.io/docs/specification/about/.

geoffcorey:
http://localhost:3333/api-docs/

Resolves #77


mckaymic:
Is there any reason to provide the razee-org-key for the API docs?

PR #89, opened 2019-10-07
Update rendered YAML to use new name 'deploy.razee.io'

mindjiver:
Generate a resource YAML file with the new renamed apiVersion
otherwise the resource will not apply cleanly on a cluster that
only has the new CRDs (Custom Resource Definitions) installed.


mindjiver:
Nevermind, I saw there was an existing one.

PR #125, opened 2020-01-14
remove orgAdminKey from get/post resutls

dalehille:
https://github.com/razee-io/Razee/issues/75

### DIFF
--- a/app/routes/v2/orgs.js
+++ b/app/routes/v2/orgs.js
@@ -41,23 +41,20 @@ const createOrg = async(req, res) => {
     return res.status(400).send( 'This org already exists' );
   }
 
-  const orgAdminKey = req.orgAdminKey; // this was set in verifyAdminOrgKey()
   const orgApiKey = `orgApiKey-${uuid()}`;
   try {
     const insertedOrg = await Orgs.insertOne({
       '_id': uuid(),
       'name': orgName,
       'orgKeys' : [ orgApiKey ],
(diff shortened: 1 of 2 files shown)


app/routes/v2/orgs.js
```diff
@@ -54,9 +54,10 @@ const createOrg = async(req, res) => {
     });
 
     if(insertedOrg.result.ok) {
+      delete insertedOrg.ops[0].orgAdminKey;
       return res.status(200).send( insertedOrg.ops[0] );
     } else {
-      req.log.error(insertedOrg);
+      req.log.error(orgName);
```
adamkingit: I would log the orgName as well as the result status, ensuring that they keys aren't included. With this change as is, we'll know what org we failed to insert, but not what it failed.

app/routes/v2/orgs.js
```diff
@@ -41,22 +41,20 @@ const createOrg = async(req, res) => {
...
       'created': new Date(),
       'updated': new Date()
     });
-
+   
     if(insertedOrg.result.ok) {
       return res.status(200).send( insertedOrg.ops[0] );
     } else {
-      req.log.error(insertedOrg);
+      req.log.error(`${res.statusCode}  Could not create ${orgName}`);
```
adamkingit: Should have made this suggestion more explicitly the first time. I think you want 
req.log.error(insertedOrg.result, `Failed to insert org ${orgname} document into Orgs collection.`);

PR #167, opened 2020-02-18
added ADD_CLUSTER_WEBHOOK_HEADERS variable which expects a json objec…

rmgraham:
…t of headers to send


rmgraham:
Implements https://github.com/razee-io/Razee/issues/80

PR #142, opened 2020-01-30
graphql (see the latest in graphql branch or this PR #173)

yingwang-us:
Do not merge this PR, see the latest in graphql branch or this https://github.com/razee-io/Razeedash-api/pull/173 PR.

PR #173, opened 2020-02-21
add apollo graphql to graphql branch (#159)

yingwang-us:
Following changes are made based on Mike's code review comments.

1. Add ENABLE_GRAPHQL environment variable , default is false. see [here](https://github.com/razee-io/Razeedash-api/blob/graphql/app/index.js#L87) and [here](https://github.com/razee-io/Razeedash-api/blob/graphql/app/routes/v2/clusters.js#L243)

2. Add req_id to graphql context, which is then used by logger to print debug information with the req_id. see [here](https://github.com/razee-io/Razeedash-api/blob/graphql/app/apollo/index.js#L54) for setting and [here](https://github.com/razee-io/Razeedash-api/blob/graphql/app/apollo/resolvers/resource.js#L43) for example how to use. Note, for web socket connections, the req_id on the context is the original http upgrade request `req.id`.

3. Add apollo's bunyan config [here](https://github.com/razee-io/Razeedash-api/blob/graphql/app/apollo/utils/bunyan.js) . This is because graphql has common url and we need to turn on body log, so we could know what graphql api is called. Also include some optimization to disable debug logs for playground ping requests.

4. Add [identifier](https://github.com/razee-io/Razeedash-api/blob/graphql/app/apollo/schema/user.js#L56) in user schema and make email as optional. It is because some tokens might not be associated with any email, e.g. a token generated from service-id api key.

5. other minor changes mentioned in Mike's code review under 
https://github.com/razee-io/Razeedash-api/pull/142


package.json
```diff
@@ -21,10 +25,15 @@
   "author": "",
   "contributors": [],
   "dependencies": {
+    "apollo-server": "^2.10.1",
+    "apollo-server-express": "^2.10.1",
```
mckaymic: Sorry @yingwang-us .  One last issue.  Why not use apollo-server-express to integrate express and apollo?  According to this https://www.apollographql.com/docs/apollo-server/v1/servers/express/ we can combine the 2 and only expose 1 port.

yingwang-us: @mckaymic combined the 2, the changes bellow are a little complicated, due to different auth, logging requirements between ingestion api and query api. But it works under port 3333.

yingwang-us:
temporarily turned off vulnerability 1217 for audit since it does not impact prod image. Still waiting https://github.com/kevva/decompress/issues/71 to be fixed.

Dockerfile
```diff
@@ -25,4 +26,5 @@ WORKDIR /usr/src/
 COPY --from=buildImg /usr/src /usr/src
 
 EXPOSE 3333
+EXPOSE 8000
```
mckaymic: Can this be removed now?

yingwang-us: nice catch, removed.

PR #201, opened 2020-03-05
add redis url and cert support

yingwang-us:
1. support url style redis config
2. self signed redis server cert
3. temporarily turn off audit on acorn package until there is a fix later

### DIFF
--- a/app/utils/pubsub.js
+++ b/app/utils/pubsub.js
@@ -76,7 +76,7 @@ var getNewClient = ()=>{
       logger.debug(`Skip loading self-signed redis cert from: ${redisCertPath}`);
     }
   } catch (err) {
-    logger.warn(err, `Ignore the redis server cert error.`);
+    logger.warn(err, 'Ignore the redis server cert error.');
   }
 
   // process redis url if the env variable is defined
--- a/app/utils/pubsub.js
+++ b/app/utils/pubsub.js
@@ -16,9 +16,13 @@
  */
 
 const fs = require('fs');
+const bunyan = require('bunyan');
 var _ = require('lodash');
 var Redis = require('ioredis');
-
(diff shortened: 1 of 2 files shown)


app/utils/pubsub.js
```diff
@@ -49,8 +50,16 @@ var init = ()=>{
 };
 
 var getNewClient = ()=>{
-  var conf = JSON.parse(process.env.REDIS_CONN_JSON || '{}');
-  return new Redis(conf);
+  var options = JSON.parse(process.env.REDIS_CONN_JSON || '{}');
```
adamkingit: Handle parse error

app/utils/pubsub.js
```diff
@@ -49,8 +50,16 @@ var init = ()=>{
...
+  if (process.env.REDIS_CERTIFICATE_PATH) {
```
adamkingit: consider just checking for the existence of the cert in the mounted config, rather than creating an env var to define where to read from

app/utils/pubsub.js
```diff
@@ -49,8 +50,16 @@ var init = ()=>{
...
+    // for self signed cert of redis server
+    options.tls = { ca: [fs.readFileSync(process.env.REDIS_CERTIFICATE_PATH)] } ;
```
adamkingit: Handle read error, or not found error. Check existence then parse will be cleaner

app/utils/pubsub.js
yingwang-us: fixed.

app/utils/pubsub.js
yingwang-us: fixed.

adamkingit:
This looks good to me. I will defer to @rmgraham for final merge approval who knows a bit more about Redis.

PR #247, opened 2020-03-30
Prometheus2

harkotha2:


app/index.js
```diff
@@ -111,3 +117,23 @@ function onError(error) {
...
+  });
+}
+
+//Prometheus server configuration
+var metrics_server = http.createServer(function (request, response) {
+  response.writeHead(200, {'Content-Type': promClient.register.contentType});
+  response.end(promClient.register.metrics());
+});
+
+metrics_server.listen(9095, () => {
```
mckaymic, line 135: This should not be a separate server.  It should just be an endpoint on the existing Express app we have already.

harkotha2: Removed the prometheus server, added a /metrics endpoint on graphql server. listen on same 3333 port.  http://localhost:3333/metrics

PR #253, opened 2020-04-01
Update NPM packages

mckaymic:


mckaymic:
I've run through a handful of tests.  I think it ready to be tested on app.razee.io

PR #254, opened 2020-04-01
push registrationUrl impl down to model

yingwang-us:
1. allow organization impl to do additional tasks when the registrationUrl api is invoked.
2. refactor and simplify context parameters, so that auth model could easily access req_id, models, logger and me(the user) from the context parameter
3. update README with instructions on how to enable graphql for development.


app/apollo/models/organization.local.schema.js
```diff
@@ -45,6 +45,19 @@ const OrganizationLocalSchema = new mongoose.Schema({
...
+OrganizationLocalSchema.statics.getRegistrationUrl = async function(org_id, context) {
+  context.logger.debug({org_id}, 'getRegistrationUrl enter');
+  const org = await this.findById(org_id);
+  if (process.env.EXTERNAL_URL) {
+    return {
+      url: `${process.env.EXTERNAL_URL}/api/install/razeedeploy-job?orgKey=${org.orgKeys[0]}`,
+    };
+  }
+  return {
+    url: `http://localhost:3333/api/install/razeedeploy-job?orgKey=${org.orgKeys[0]}`,
```
mckaymic, line 57: Is there a way to get the real address instead of hardcoding it to localhost?

yingwang-us: fixed and removed EXTERNAL_URL env.

PR #274, opened 2020-04-09
Added the default schema for orgs

mckaymic:


app/apollo/models/organization.default.schema.js
```diff
@@ -0,0 +1,70 @@
...
+  context.logger.debug({org_id}, 'getRegistrationUrl enter');
+  const org = await this.findById(org_id);
+  const protocol = context.req ? context.req.protocol : 'http';
+  const host = context.req ? context.req.header('host') : 'localhost:3333';
+  return {
+    url: `${protocol}://${host}/api/install/razeedeploy-job?orgKey=${org.orgKeys[0]}`,
+  }; 
+};
+
+OrganizationLocalSchema.statics.createLocalOrg = async function(args) {
```
yingwang-us, line 57: @mckaymic createLocalOrg method for local/passport.local auth model is only used for unit/api test purpose, so it may not be required here.

PR #288, opened 2020-04-14
Chan sub UUID

rmgraham:
Makes the code use the `uuid` field as the identifier for chans/subs rather than _id. Also adds some pubsub publish calls, although websockets in razeeapi look broken right now, so I cant test them.

`mks-ui/pull/57` aligns with these changes, so ideally we'd get them both pushed out at the same time.

### DIFF
--- a/app/apollo/resolvers/channel.js
+++ b/app/apollo/resolvers/channel.js
@@ -15,7 +15,7 @@
  */
 
 const _ = require('lodash');
-const uuid = require('uuid').v4;
+const UUID = require('uuid').v4;
 const crypto = require('crypto');
 
 const { ACTIONS, TYPES } = require('../models/const');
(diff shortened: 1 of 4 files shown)


app/apollo/resolvers/channel.js
```diff
@@ -15,7 +15,7 @@
  */
 
 const _ = require('lodash');
-const uuid = require('uuid').v4;
+const UUID = require('uuid').v4;
```
mckaymic, line 18: Needs to be updated to this format `const { v4: uuid } = require('uuid');`. The current format is deprecated and throws warnings now

app/apollo/resolvers/channel.js
```diff
@@ -49,35 +49,36 @@ const resourceResolvers = {
       await validAuth(me, org_id, ACTIONS.MANAGE, TYPES.CHANNEL, queryName, context);
 
       try{
-        const _id = uuid();
+        const uuid = UUID();
 
         await models.Channel.create({
-          _id, org_id, name, uuid: uuid(), versions: [],
+          _id: UUID(),
```
mckaymic, line 55: Can we use the same `uuid` here?

app/apollo/resolvers/channel.js
```diff
@@ -165,9 +166,9 @@ const resourceResolvers = {
       const data = await encryptOrgData(orgKey, content);
 
       const deployableVersionObj = {
-        _id: uuid(),
-        org_id: org_id,
-        uuid: uuid(),
+        _id: UUID(),
```
mckaymic, line 169: Should we make these the same also?

app/apollo/resolvers/subscription.js
```diff
@@ -15,7 +15,8 @@
  */
 
 const _ = require('lodash');
-const uuid = require('uuid').v4;
+const UUID = require('uuid').v4;
```
mckaymic, line 18: `const { v4: uuid } = require('uuid');`

app/apollo/resolvers/subscription.js
```diff
@@ -88,28 +89,36 @@ const resourceResolvers = {
         }
 
         await models.Subscription.create({
-          _id, org_id, name, tags, uuid: uuid(), owner: me._id,
+          _id: UUID(),
```
mckaymic, line 92: Same comment, should they be the same?

PR #321, opened 2020-05-05
apollo-metrics enablement

pawan-chawdhary:


app/apollo/index.js
```diff
@@ -72,6 +75,8 @@ const buildCommonApolloContext = async ({ models, req, res, connection, logger }
 const createApolloServer = () => {
   const server = new ApolloServer({
     introspection: true,
+    plugins: [apolloMetricsPlugin],
```
yingwang-us, line 78: could we also add process.env.GRAPHQL_ENABLE_TRACING === 'true' here to pass empty plugin [] when it is not enabled?

pawan-chawdhary: updated

PR #334, opened 2020-05-11
Cluster subscription gql user token

rmgraham:
Lets you do a userToken for auth:
image

Obj looks like this:
image

And allows any of the org's orgKeys as the secret

### DIFF
--- a/app/apollo/models/user.js
+++ b/app/apollo/models/user.js
@@ -23,36 +23,9 @@ const UserSchema = require(`./user.${AUTH_MODEL}.schema`);
 const _ = require('lodash');
 
 const loadMeFromUserToken = async function(userToken, models){
-  let obj, userId, orgId;
-  try {
-    obj = jwt.decode(userToken);
-    userId = obj.userId;
-    orgId = obj.orgId;
-  }catch(err){
-    throw new AuthenticationError('Failed to parse userToken');
-  }
-  if(!userId){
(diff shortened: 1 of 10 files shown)


app/apollo/models/user.js
```diff
@@ -15,11 +15,75 @@
...
+    throw new AuthenticationError('userToken could not be verified');
+  }
+  return {
+    type: 'userToken',
+    user,
+  };
+};
+
+const getMeFromConnectionParamsBase = UserSchema.statics.getMeFromConnectionParams;
+UserSchema.statics.getMeFromRequest = async function(...args){
```
yingwang-us, line 64: is this line should be `UserSchema.statics.getMeFromConnectionParams = .....` ?

app/apollo/models/user.js
```diff
@@ -15,11 +15,75 @@
...
+  if(!user){
+    throw new AuthenticationError('No user found for userToken');
+  }
+  const org = await models.Organization.findById(orgId);
+  if(!org){
+    throw new AuthenticationError('No org found for userToken');
+  }
+  const hasVerifiedToken = _.some(org.orgKeys, (orgKey)=>{
+    try{
+      jwt.verify(userToken, orgKey);
```
yingwang-us, line 47: for the `secretKey`, could we let it = orgKey + some static secret ? That way even some users of the org may know orgKey, they are not be able to sign and pass verify for userToken.

PR #365, opened 2020-05-28
added registerCluster API

johnymathew:
https://github.com/razee-io/Razee/issues/105

see razeedeploy-delta change here: https://github.com/razee-io/razeedeploy-delta/pull/103


yingwang-us:
after the watch-keeper reported back, the db entry looks like:

image

yingwang-us:
@mike, updated with 404 watch-keeper response if a cluster is not registered and default to REGISTERING state if reg_state is not defined. Still like put required field 1name1 in registration JSON parameter since it looks like more clean. but if you want, I could add additional `name` parameter explicitly.

PR #372, opened 2020-06-03
Auto delete the soft-deleted resources from db

harkotha2:


app/apollo/resolvers/cluster.js
```diff
@@ -203,7 +203,7 @@ const clusterResolvers = {
         //TODO: soft delete the resources for now. We need to have a background process to
         // clean up S3 contents based on deleted flag. 
         const deletedResources = await models.Resource.updateMany({ org_id, cluster_id }, 
-          {$set: { deleted: true }}, { upsert: false });
+          {$set: { deleted: true, expireAt: new Date() }}, { upsert: false });
```
mckaymic, line 206: I think https://docs.mongodb.com/manual/reference/operator/update/currentDate/ might be a better option.  Just in case the database and server process are not in sync with time

harkotha2: Modified date format(`new Date()`) to `$currentDate`. Please check

mckaymic:
Not going to use this model

PR #386, opened 2020-06-08
made resources search endpoints to return { count: int, resources: [R…

rmgraham:
made resources search endpoints return:
{ count: int, resources: [Resource!]!}
instead of just [Resource!]

Also need to coordinate with ui team, since this is a breaking api change


app/apollo/resolvers/resource.js
```diff
@@ -35,8 +35,9 @@ const commonResourcesSearch = async ({ models, org_id, searchFilter, limit, quer
       .limit(limit)
       .lean()
     ;
+    var count = await models.Resource.find(searchFilter).count();
```
dalehille, line 38: would we be able to just do `count = resources.length` rather than making a second call into the Resources collection?

rmgraham: The returned results are what we have after the limit is applied, so theyre capped at a certain number. For `count` we want to see the total number of matches in the db.

PR #395, opened 2020-06-09
New: add new functions in channel

worksmartAmy:


app/apollo/resolvers/channel.js
```diff
@@ -271,7 +271,29 @@ const channelResolvers = {
         version_uuid: versionObj.uuid,
       };
     },
-
+    uploadChannelVersion: async (parent, { org_id, uuid, content }, context)=>{
```
yingwang-us, line 274: we need to support S3. see here: https://github.com/razee-io/Razeedash-api/blob/master/app/apollo/resolvers/channel.js#L212 for details.

app/apollo/resolvers/channel.js
```diff
@@ -302,6 +324,42 @@ const channelResolvers = {
         throw err;
       }
     },
+    removeChannelVersion: async (parent, { org_id, uuid }, context)=>{
```
yingwang-us, line 327: we need to support S3. see here: https://github.com/razee-io/Razeedash-api/blob/master/app/apollo/resolvers/channel.js#L212 for details.

PR #403, opened 2020-06-10
New: Label implementation part I

yingwang-us:
Design:
- Label is an immutable entry in label db with tuple  values `{orgId, name, uuid, owner}`. That means `name` and `uuid` will not be changed, unless you delete the entry.
- Query Label APIs
   - `labels(orgId: String!): [Label]`
   - `label(orgId: String! uuid: String!): LabelDetail`
   - `labelByName(orgId: String! name: String!): LabelDetail`
- Mutation APIs
   - `addLabel(orgId: String! name: String!): AddLabelReply!`
   - `removeLabel(orgId: String! uuid: String!): RemoveLabelReply!`
   - `removeLabelByName(orgId: String! name: String!): RemoveLabelReply!`
   - `labelClusters(orgId: String! uuid: String! clusters: [String]!): LabelClustersReply!`
   - `unlabelClusters(orgId: String! uuid: String! clusters: [String]!): UnlabelClustersReply!`
- Impacts/migration on subscription api
   - all subscription api continue using tags (a.k.a. label.name)
   - before env variable `LABEL_VALIDATION_REQUIRED` is enabled, all existing tags for subscriptions will be populated into label database either from user-facing APIs or from subscriptionsByTag. if populated from subscriptionsByTag, owner will be set to undefined
   - if env variable `LABEL_VALIDATION_REQUIRED` is enabled, `subscriptionsByTag` will throw exception, and users should migrate clustersubscription crd to use 
       - `subscriptionsByCluster(cluster_id: String): [UpdatedSubscription]`
- Impacts on cluster api
   - `cluster.tags` are exposed from list/get cluster APIs.
   - since existing clustersubscrition agent does not send cluster_id, we can not migrate and populate tags to the clusters db.
- Label based auth against sub/cluster/resource APIs will be in part II
- Misc: fixes to include cluster-facing sub api for local, passport.local auth model


dalehille:
can you update the description to match https://github.com/razee-io/Razee/blob/master/commit-message-guidelines.md ?

yingwang-us:
there will be a follow-up clusterSubscription change to use `subscriptionsByCluster(cluster_id: String): [UpdatedSubscription]`. I might need help from @dalehille  on that.

PR #418, opened 2020-06-15
New: Resource history

rmgraham:
Adds `resourceHistory`, which shows the edit timestamps and _id's for resource yaml history.
Also added $histId as an optional arg to `resource`, which tries to load the `yamlStr` and `updated` time based on the resourceYamlHist obj with that id


app/apollo/schema/channel.js
```diff
@@ -83,6 +87,11 @@ const channelSchema = gql`
      Gets a yaml version from this channel
      """
      getChannelVersion(org_id: String!, channel_uuid: String!, version_uuid: String!): DeployableVersion!
+     
+     """
+     Lets the $limit most recent versions for a channel
+     """
+     getChannelVersions(org_id: String!, channel_uuid: String!, limit: Int = 500): DeployableVersionList!
```
dalehille, line 94: I don't see the implementation of getChannelVersions in resolvers/channels.js

dalehille:

PR #423, opened 2020-06-16
New: Label implementation part II - label based auth proposal

yingwang-us:
Design:

Most auth models use OR logic including IAM. e.g. if a user is granted to access resources for `prod` and resources for `us-south`, that normally means the user is granted to access resources from both sets, NOT the intersection of 2 sets. 

We decided to use `AND` logic for clusters/resources/subscriptions auth. This means that User's tag set >= targeted clusters/resources/subscriptions tag sets. It also means that Users with `prod` tag may not see any resources, unless he has tags from other dimensions too. 

e.g. 3 dimensions [ envs: {'prod', 'stage'}, regions: {'us-east', 'eu-de' },  apps: {'edge', 'database'} ] tag system. A user needs at least one of each dimension in order to see something. To allow a users to see all `prod` resources he needs following to have permission on `[prod, us-east, eu-de, edge, database ]` labels. Ideally an auth model could support dimentional `[prod, *, * ]`, but for IAM we only has `resourceType(used by label):resourceId`

Implementation:

- `isAuthorizedBatch = async function(me, orgId, objectArray, context)` , where objectArray are list of {`action`, `type`, `uuid`} and the return is an array of [true|false] . This should be implemented by each auth model.

- `getUserTags = async (me, org_id, action, field, queryName, context)` return a list of tags' uuids or names that user is permitted for the `action`.

- `getUserTagConditions = async (me, org_id, action, field, queryName, context)` return a mongo filter that all clusters/subscriptions tags must exist in user tags.

- `getUserTagConditionsIncludingEmpty = async (me, org_id, action, field, queryName, context)` return a mongo filter that all clusters/subscriptions tags must exist in user tags or all clusters/subscriptions tags are empty. 

- `resources (org_id, filter, fromDate, toDate, limit, kinds = [ ], sort)` requires `servicelevelread` action, and mapped to ADMIN role for local and coresponding user data are updated. The action will be mapped to proper role for IAM.

- remove unnecessary check on `if (AUTH_MODEL === AUTH_MODELS.LOCAL)`


dalehille:
@yingwang-us in the addLabel mutation in this section:

```await models.Label.create({
          _id: UUID(),
          uuid, orgId, name, owner: me._id,
        });
```
 `owner: me._id` will be undefined for razeedash users.  Do you know how we can get me._id populated for `userToken` users

yingwang-us:
@dalehille in `getMeFromRequest` or `getMeFromConnection` methods. In the functions, the token will be verified, user information should be decoded back from token directly or make another db lookup based on the info from the token.

PR #458, opened 2020-06-29
New: Update graphql variables to camelCase

harkotha2:
In this PR, we updated all graphql methods, parameters, output varible names to camel-case. Note: values in JSON fields might not be updated.

Note: There is no impact on db, we continue using org_id, cluster_id, etc in mongo db. The changes are done purely by utilizing nodejs alias and mongoose alias. 

Temporarily whitelist https://www.npmjs.com/advisories/1523 until there is a fix merged into lodash https://github.com/lodash/lodash/pull/4759


mckaymic:
We need to make sure we give the core and UI teams a heads up before we deploy this change.

PR #486, opened 2020-07-14
New: Made mongoose throw an error if we're saving an object that doesnt pass the schema

rmgraham:
Previously it would silently ignore the invalid fields


app/apollo/resolvers/subscription.js
```diff
@@ -257,9 +257,9 @@ const subscriptionResolvers = {
           name, groups,
           channelName: channel.name, channel_uuid, version: version.name, version_uuid,
         };
-        await models.Subscription.updateOne({ uuid, org_id, }, { $set: sets });
+        await models.Subscription.updateOne({ uuid, org_id: orgId, }, { $set: sets });
 
-        pubSub.channelSubChangedFunc({org_id: org_id});
+        pubSub.channelSubChangedFunc({ orgId });
```
dalehille, line 262: I think this needs to remain as `{org_id: org_id}` since inside `channelSubChangedFunc` we reference `data.org_id`

PR #495, opened 2020-07-17
Bug: missing await and sub for some resource(s) apis

yingwang-us:
1. bug fix: missing await causes missing fields and lean is not working.

2. render subscriptions for resources if users ask for them.

3. fix `resource` API. Set `resource._id` type to `ObjectId` because it is populated by REST API as ObjectId type. which is caused by the change: https://github.com/razee-io/Razeedash-api/commit/d55006eb68b7256bde83f2f84027811d01102712#diff-2b03232d3d71f02aa4120c9e3d2c7befR21

4. quick fix groups api data.

### DIFF
--- a/app/apollo/models/resource.schema.js
+++ b/app/apollo/models/resource.schema.js
@@ -15,10 +15,11 @@
  */
 
 const mongoose = require('mongoose');
+const ObjectId = require('mongoose').Types.ObjectId;
 
 const ResourceSchema = new mongoose.Schema({
   _id: {
-    type: String,
+    type: ObjectId,
     alias: 'id',
   },
   org_id: {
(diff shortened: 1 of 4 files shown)


app/apollo/models/resource.schema.js
```diff
@@ -15,10 +15,11 @@
  */
 
 const mongoose = require('mongoose');
+const ObjectId = require('mongoose').Types.ObjectId;
 
 const ResourceSchema = new mongoose.Schema({
   _id: {
-    type: String,
+    type: ObjectId,
```
rmgraham, line 22: Wont forcing this to be an objectid cause problems later? Because we'll need to update all the foreign keys in other tables to be objectids too, and objectids are also effort to compare.

yingwang-us: we probably do that later. this is a quick patch to get staging working.

PR #558, opened 2020-08-05
Set Upload yaml file size limit & parse yaml content

harkotha2:


app/apollo/models/const.js
```diff
@@ -19,6 +19,7 @@ const {ACTIONS, TYPES, AUTH_MODELS, AUTH_MODEL } = require('../../utils/auth.con
 const SECRET = process.env.SECRET || 'very-very-secret';
 const GRAPHQL_PATH = process.env.GRAPHQL_PATH || '/graphql';
 const APOLLO_STREAM_SHARDING = process.env.APOLLO_STREAM_SHARDING === 'false' ? false : true;
+const CHANNEL_VERSION_YAML_MAX_SIZE_LIMIT = process.env.CHANNEL_VERSION_YAML_MAX_SIZE_LIMIT || 30;
```
mckaymic, line 22: Can we add the units in the env variable (i.e. MB).  Otherwise it's hard to remember what units to use.

harkotha2: Added comment in the file

harkotha2: Added units(MB) in the variable - `CHANNEL_VERSION_YAML_MAX_SIZE_LIMIT_MB`

PR #562, opened 2020-08-06
New: Multi nested attrs

rmgraham:
Automatically populates child attributes when requested. i.e. going from group -> clusters -> resources. This is so we dont have to write duplicate code any time we have an endpoint we want to load child attributes from.

Allows for queries with multi-nested children like this:
image
(group -> clusters -> resources -> cluster -> resources). It doesnt make sense that someone would ask for that, but its still possible.


app/apollo/resolvers/resource.js
```diff
@@ -21,15 +21,16 @@ const GraphqlFields = require('graphql-fields');
...
 const ObjectId = require('mongoose').Types.ObjectId;
+const { applyQueryFieldsToResources } = require('../utils/applyQueryFields');
 
 const conf = require('../../conf.js').conf;
 const S3ClientClass = require('../../s3/s3Client');
 const url = require('url');
 
 // This is service level search function which does not verify user tag permission
-const commonResourcesSearch = async ({ context, org_id, searchFilter, limit=500, queryFields, sort={created: -1} }) => {
+const commonResourcesSearch = async ({ context, searchFilter, limit=500, queryFields, sort={created: -1} }) => { // eslint-disable-line
```
dalehille, line 33: the places in the code that are calling `commonResourcesSearch` are still passing in org_id.  Should those be updated to no longer do that?

app/apollo/resolvers/resource.js
```diff
@@ -39,9 +40,23 @@ const commonResourcesSearch = async ({ context, org_id, searchFilter, limit=500,
     ;
     var count = await models.Resource.find(searchFilter).count();
     // if user is requesting the cluster field (i.e cluster_id/name), then adds it to the results
-    if( (queryFields.resources||{}).cluster ) {
-      await applyClusterInfoOnResources(org_id, resources, models);
-    }
+    // if(queryFields.cluster){
```
dalehille, line 43: do we need the comments?

PR #563, opened 2020-08-06
New: remove deprecated and un-used API

worksmartAmy:
Removed clusterDistributed and resourceDistributed
Removed organization in organization resolver
Renamed clusterZombies to inactiveClusters


app/apollo/index.js
```diff
@@ -27,7 +27,7 @@ const { AUTH_MODEL, GRAPHQL_PATH } = require('./models/const');
 const typeDefs = require('./schema');
 const resolvers = require('./resolvers');
 const recoveryHintsMap = require('./resolvers/recoveryHintsMap');
-const { models, connectDb, setupDistributedCollections, closeDistributedConnections } = require('./models');
+const { models, connectDb, /*setupDistributedCollections, closeDistributedConnections*/ } = require('./models');
```
yingwang-us, line 30: Hi Amy, could we just remove the fields or lines instead of commenting them? if we want we could easy cover them back. Same for other files. Other than that, the PR looks good.

worksmartAmy: done

PR #590, opened 2020-08-14
Update: Add more retriction for string type

yingwang-us:
- [x] Add more retriction for string type
- [x] Validate json key and value as well


yingwang-us:
will close this pr since it is a dup of Johny's pr.

PR #616, opened 2020-08-20
Add init pubsub retries & simple live test & simple apollo health check

yingwang-us:
- [x] add retries for initial redis pub/sub setup
- [x] add simple liveness check for pub/sub. A complete runtime pubsub probe test will be added later
- [x] add custom apollo health check for `/.well-known/apollo/server-health` on apollo side


mckaymic:
@yingwang-us We should also update the app/apollo/subscription/index.js so that the `channelSubChangedFunc` and `resourceChangedFunc` throw an error rather than just logging a message if the Redis post fails

PR #620, opened 2020-08-22
Allow playground auth middleware

yingwang-us:
Allow playground auth middleware route, if it is NOT production environment and playgroundAuth is defined.


yingwang-us:
@mckaymic we will use the same flag to control both playground and introspection. For playground it is just a simple GET request, it is easy to use the same flag to control both if environment is not production env.

PR #626, opened 2020-08-24
add liveness check for pubsub

worksmartAmy:


app/apollo/subscription/index.js
```diff
@@ -53,6 +53,7 @@ class PubSubImpl {
   
   constructor(params) {
     this.initRetries = 0;
+    this.pubsubQueue = new Array();
```
yingwang-us, line 56: we probably do not need an array here, as long as the variable could store one latest {timestamp: <date-time>} should be enough.

app/apollo/subscription/index.js
```diff
@@ -83,6 +84,8 @@ class PubSubImpl {
       logger.info(
         `Apollo streaming is enabled on redis endpoint ${url.hostname}:${url.port}`,
       );
+      this.livenessCheckPublisher();
+      this.livenessCheckSubscriber(this.pubsubQueue);
```
yingwang-us, line 88: we do not need parameter, you could get variable from this.variablename in the function.

app/apollo/subscription/index.js
```diff
@@ -99,7 +102,31 @@ class PubSubImpl {
     }
     return false;
   }
-
+  async livenessCheckSubscriber(pubsubQueue){
+    if (process.env.NODE_ENV !== 'unit-test' && process.env.NODE_ENV !== 'test') {
+      const sizeLimit = 10;
+      this.pubSub.redisSubscriber.on('message', function(channel, message){
+        //console.log('receiving: ', JSON.parse(message));
```
yingwang-us, line 109: we need to filter channel here since we only want to store the test data for channel == 'testTopic'

PR #656, opened 2020-09-28
pass the channel UUID as the attribute

keithmorganIBM:
this should create a CRN that looks like:
`crn:v1:bluemix:public:satellite:global:a/{orgId}::configuration:{channel_uuid}`

altho, to really make this work, we may need to change it to pass the channel name:
`crn:v1:bluemix:public:satellite:global:a/{orgId}::configuration:{channel_name}`

...but that's weird.


mckaymic:
@keithmorganIBM It looks like the IAM module in Satellite-Config already supports the attributes, so this should work with it.  https://github.ibm.com/alchemy-containers/satellite-config/blob/master/Sloop-api/app/apollo/iamutils/iamWrapperNew.js

rmgraham (approved):

PR #667, opened 2020-10-09
New: filtered channels list to what you have access to

rmgraham:


app/apollo/resolvers/common.js
```diff
@@ -38,7 +38,30 @@ const validClusterAuth = async (me, queryName, context) => {
     }
     return;
   }
-}; 
+};
+
+var getAllowedChannels = async(me, orgId, action, field, context)=>{
```
dalehille, line 43: `field` is passed in to `getAllowedChannels` and `filterChannelsToAllowed` but it doesn't seem to be in use in those functions

dalehille (approved):

PR #679, opened 2020-10-13
namespace filtering for resources

keithmorganIBM:
round one of namespace filtering.  will test soon...but its pretty straightforward.


app/apollo/resolvers/resource.js
```diff
@@ -29,6 +29,33 @@ const conf = require('../../conf.js').conf;
...
+  const deleteArray = [];
+  const namespaces = res.map(d => d.searchableData.namespace).filter((v, i, a) => a.indexOf(v) === i).filter(x => x);
+
+  //need to adjust validateAuth to retun false and not throw error
+  namespaces.forEach(n => {
+    async () => {
+      const valid = await validAuth(me, orgId, ACTIONS.READ, TYPES.RESOURCE, queryName, context, [n]);
+      if (!valid) deleteArray.push(n);
+    };
+  });
```
rmgraham, line 44: js foreach's dont run `await` on the called function. So `deleteArray` will always be `[]` at the moment the foreach finishes.

`await Promise.all(namespaces.map(async()=>{ ... }))` would work for the most part.

Although we might want to put concurrency limits on it, in case we have 1000 namespaces. Other parts of the code use the `p-limit` library for that.

keithmorganIBM: 🥇  for the promise suggestion.  I really didnt want to put in that foreach.

keithmorganIBM:
https://github.ibm.com/alchemy-containers/satellite-planning/issues/927

rmgraham (approved):

PR #680, opened 2020-10-13
Update : Removed console.log(`user01 token=${token}`) 

tasneem067:
Removed "console.log(`user01 token=${token}`)"


app/apollo/test/resource.spec.js
```diff
@@ -683,7 +683,7 @@ describe('resource graphql test suite', () => {
           }
         }`;
 
-        const meResult = await api.me(token);
+        const meResult = await api.me();
```
dalehille, line 686: this line should remain as `const meResult = await api.me(token);`

dalehille (changes requested):

dalehille (approved):

PR #772, opened 2021-02-04
feat:Translation files

tasneem067:


app/index.js
```diff
@@ -64,7 +66,20 @@ app.use(function errorHandler(err, req, res, next) {
   }
   next();
 });
-
+i18next.use(Backend).use(i18nextMiddleware.LanguageDetector).init({
+  backend: {
+    loadPath:'./locales/{{lng}}/razee-resources.json',
+    addPath:'./locales/{{lng}}/{{ns}}.missing.json'
```
rmgraham, line 72: This generates a translation.missing.json file. Which we then never use.
Test it by putting `context.req.t('asdf')` somewhere in the code, running it, and seeing en/translation.missing.json. And then running `npm run i18n`, and still seeing the missing.json file still there.

app/apollo/index.js
```diff
@@ -38,6 +38,24 @@ const initModule = require(`./init.${AUTH_MODEL}`);
...
+const i18nextMiddleware = require('i18next-http-middleware');
+const Backend = require('i18next-fs-backend');
+i18next.use(Backend).use(i18nextMiddleware.LanguageDetector).init({
+  //debug: true,
+  backend: {
+    loadPath: './locales/{{lng}}/razee-resources.json',
+    addPath: './locales/{{lng}}/{{ns}}.missing.json'
+  },
+  fallbackLng: 'en',
+  supportedLngs: ['en', 'cn', 'zh-cn'],
```
rmgraham, line 51: Why does this one have a few langs and the other one a bunch?

app/index.js
tasneem067: yes,missing.json file is created when there is a missing string.

rmgraham: Yea, but we dont use that file. It just gets in the way, since `npm run i18n` populates the other json files properly and lets missing.json sit around

rmgraham (approved):

PR #778, opened 2021-02-09
added clusterId as an option to subscriptions in the place of groups.…

rmgraham:
Added clusterId as an option to subscriptions in the place of groups.
Used when users want to assign a sub to a single specific cluster and not be bothered by creating groups.

### DIFF
--- a/app/apollo/schema/subscription.js
+++ b/app/apollo/schema/subscription.js
@@ -113,12 +113,12 @@ const subscriptionSchema = gql`
      """
      Adds a subscription
      """
-     addSubscription(orgId: String! @sv, name: String! @sv, groups: [String!] @sv, channelUuid: String! @sv, versionUuid: String! @sv, clusterId: String): AddChannelSubscriptionReply!
+     addSubscription(orgId: String! @sv, name: String! @sv, groups: [String!] @sv, channelUuid: String! @sv, versionUuid: String! @sv, clusterId: String @sv): AddChannelSubscriptionReply!
      
      """
      Edits a subscription
      """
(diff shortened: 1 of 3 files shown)


app/apollo/schema/subscription.js
```diff
@@ -112,12 +113,12 @@ const subscriptionSchema = gql`
      """
      Adds a subscription
      """
-     addSubscription(orgId: String! @sv, name: String! @sv, groups: [String!]! @sv, channelUuid: String! @sv, versionUuid: String! @sv): AddChannelSubscriptionReply!
+     addSubscription(orgId: String! @sv, name: String! @sv, groups: [String!] @sv, channelUuid: String! @sv, versionUuid: String! @sv, clusterId: String): AddChannelSubscriptionReply!
```
dalehille, line 116: does the string validation ( @sv ) function need to be applied to clusterId here?   same with line 121 for editSubscription

dalehille (approved):

app/apollo/schema/subscription.js
```diff
@@ -112,12 +113,12 @@ const subscriptionSchema = gql`
      """
      Adds a subscription
      """
-     addSubscription(orgId: String! @sv, name: String! @sv, groups: [String!]! @sv, channelUuid: String! @sv, versionUuid: String! @sv): AddChannelSubscriptionReply!
```
boddumanohar, line 115: why was this `!` removed?

rmgraham: It defaults to a blank array in the code if unset. Its so if someone wants to based the subscription off a clusterId, they dont need to pass the groups at all.

PR #808, opened 2021-03-09
Update readiness to rely on apollo health, extended with SIGTERM check

adamkingit:


adamkingit:
Pretty sure this is right, but not sure how to test it locally.

app/apollo/index.js
```diff
@@ -183,6 +186,13 @@ const createApolloServer = () => {
         );
       },
     },
+    onHealthCheck: async() => {
```
dalehille, line 189: `onHealthCheck` is already defined at around line 225.  Should this logic be moved down there instead?

dalehille (changes requested):

app/apollo/index.js
adamkingit: Merged with what was there

dalehille (approved):

PR #827, opened 2021-03-18
Consolidate route logging

adamkingit:


app/utils/bunyan.js
```diff
@@ -36,14 +36,15 @@ const getBunyanConfig = (route) => {
...
-      }]      
+      }]
     };
   }
   let result = {
     name: route,
     parseUA: false,
     excludes: ['referer', 'url', 'body', 'short-body'],
     levelFn: responseCodeMapper,
+    obfuscate: ['razee-org-key'],
```
dalehille, line 47: should this be `obfuscate: ['req.headers.razee-org-key'],` instead?

adamkingit: probably/ I'll push an update for this as well as to make the config consistent

dalehille (approved):

PR #850, opened 2021-03-29
fix:Enhance apollo logging

tasneem067:


app/apollo/utils/bunyan.js
```diff
@@ -40,9 +40,9 @@ const getBunyanConfig = (route) => {
   const result = {
     name: route,
     parseUA: false,
-    excludes: ['referer', 'body', 'short-body'],
+    excludes: ['referer', 'short-body'],
     levelFn: responseCodeMapper,
-    obfuscate: ['req.headers.razee-org-key'],
+    obfuscate: ['req.headers.razee-org-key', 'req.body.variables'],
```
dalehille, line 45: We still need to see variables because that helps us understand more about a query especially when someone passes in the custom `mongoQuery` string as a variable.

dalehille (changes requested):

dalehille:
@tasneem067 can you investigate creating an apollo plugin to capture graphql request and response details?

app/apollo/utils/bunyan.js
```diff
@@ -40,9 +40,9 @@ const getBunyanConfig = (route) => {
   const result = {
     name: route,
     parseUA: false,
-    excludes: ['referer', 'body', 'short-body'],
+    excludes: ['referer', 'short-body'],
     levelFn: responseCodeMapper,
-    obfuscate: ['req.headers.razee-org-key'],
+    obfuscate: ['req.headers.razee-org-key', 'req.body.variables.login', 'req.body.variables.password', 'req.body.variables.email', 'req.body.variables.name'],
```
dalehille, line 45: x-api-key should also be in the obfuscate list

dalehille:
@tasneem067 can you make sure that when watch-keeper posts a resource to razeedash-api that we don't log the contents of the resource?

dalehille (approved):

PR #858, opened 2021-04-01
Sig int

adamkingit:


package.json
```diff
@@ -96,6 +96,7 @@
     "passport": "^0.4.1",
     "passport-local": "^1.0.0",
     "prom-client": "^13.1.0",
+    "stoppable": "^1.1.0",
```
dalehille, line 99: @adamkingit the [stoppable](https://github.com/hunterloftis/stoppable) module hasn't had an update since Oct 2019 which could be a problem if security dependencies aren't getting updated.   Interestingly, the apollo-server package also uses stoppable as a dependency..

PR #896, opened 2021-05-20
Create an index of resources by subscription_id

adamkingit:


adamkingit:
Queries such as https://github.com/razee-io/Razeedash-api/blob/master/app/apollo/utils/applyQueryFields.js#L252 do not exclude deleted fields, so I excluded deleted status though most indexes do. Should the queries include deleted: false, or should the indexes omit deleted. I haven't yet found where deleted documents get cleaned up.

dalehille (approved):

PR #909, opened 2021-06-02
Reloadable storage config and unit tests

michael-topchiev:


app/apollo/resolvers/channel.js
```diff
@@ -368,7 +368,7 @@ const channelResolvers = {
         }
 
         // deletes the linked deployableVersions in s3
-        var versionsToDeleteFromS3 = await models.DeployableVersion.find({ org_id, channel_id: channel.uuid, location: 's3', });
```
michael-topchiev, line 371: This should fix a small bug

app/apollo/models/channel.schema.js
```diff
@@ -57,9 +57,6 @@ const ChannelSchema = new mongoose.Schema({
       description: {
         type: String,
       },
-      location: {
```
carrolp, line 60: If location is actually removed from the schema, will that affect the ability to make a graphql query to get 'location'?  Code such as the CLI refers to location, even if the value is not used, e.g. https://github.ibm.com/alchemy-containers/armada-cli/blob/master/plugin/models/razee/configuration.go#L30
If this change would cause queries to fail, it breaks backwards compatibility even if the location attribute is essentially 'meaningless'.  It would be safer to keep it in the schema I would think... but if we think it's better to remove it then we need to thoroughly test in Stage, including with the existing CLI, to confirm no breakage.

.travis.yml
```diff
@@ -10,7 +10,7 @@ before_install:
 
 script:
   # Audit npm packages. Fail build whan a PR audit fails, otherwise report the vulnerability and proceed.
-  - if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then npx audit-ci --low -a node-fetch ws; else npx audit-ci --low -a node-fetch ws || true; fi
+  - if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then npx audit-ci --low -a node-fetch ws glob-parent yargs-parser; else npx audit-ci --low -a node-fetch ws glob-parent yargs-parser || true; fi
```
carrolp, line 13: Why is this change necessary?  Is it bypassing audit checks only for pull request checks or something?

app/apollo/models/channel.schema.js
michael-topchiev: mongo location field is not used anymore.

michael-topchiev: tested again - location field is still returned:
{
  "data": {
    "channels": [
      {
        "uuid": "3b49a7b4-b92f-45f5-80a1-02611ec26e4b",
        "orgId": "851d4530-75dc-4a60-9fc0-e6240316b2de",
        "name": "service-team-channel",
        "created": "2021-06-04T15:07:40.853Z",
        "data_location": "wdc",
        "versions": [
          {
            "name": "test-errors",
            "description": null,
            "uuid": "c5b53b04-1dae-4a9f-b850-a0f7f76b94bd",
            "created": "2021-06-04T15:09:02.307Z",
            "location": null
          }
        ]
      }
    ]
  }
}

.travis.yml
michael-topchiev: Could not make it pass, no safe version is available and another one requires major version library upgrade.

michael-topchiev: @carrolp please help figure out how to deal with "ws" upgrade

carrolp: graphql-tools package (that depends on the bad version of ws) not actually used, and is now removed.

carrolp (approved):

PR #911, opened 2021-06-08
make owner (aka BasicUser) objs nullable

rmgraham:


carrolp:
Not clear if we need this now -- only occurred when I directly manipulated the data in the database, not seen in production at this time.

PR #924, opened 2021-06-21
Logging

michael-topchiev:


michael-topchiev:
These changes fix logging problems at the global level, making sure every record has request ID and Ord ID, and eliminates "noise" from the logs. 

Analysis of each particular function will be needed to see if any local logging optimizations are required. Or they can be done as a result of working on production problems when we see inadequate logging in specific functional areas.

SatCon codebase will require changes according to the framework established here. Razeedash-api and SatCon will need to be launched to Staging env in tandem.

carrolp:
@michael-topchiev it looks generally as I would expect for the logging changes, except:
https://github.com/razee-io/Razeedash-api/pull/924/files#diff-f62f48f0e09390150ee44d0f9e750c0d9972f8011de86be758e8f21e558534bbL148-L158
Why is the isAuthorized function removed?  I would think that it's still used from https://github.com/razee-io/Razeedash-api/blob/master/app/utils/auth_local.js#L59 and https://github.com/razee-io/Razeedash-api/blob/master/app/apollo/resolvers/common.js#L161, right?

carrolp (approved):

PR #929, opened 2021-06-29
Add a custom attribute to 'subscription' and 'channel'

jscd:
This adds an additional field 'custom' to both subscriptions and channels, and adds changes to the GraphQL API to support the optional fields.


app/apollo/models/subscription.schema.js
```diff
@@ -72,6 +72,9 @@ const SubscriptionSchema = new mongoose.Schema({
     type: Date,
     default: Date.now,
   },
+  custom: {
+    type: String,
+  },
```
adamkingit, line 77: Thinking we want this to be an array of key/value pairs. If I am reading correctly, this would be a single string

carrolp (changes requested):
The test cases in https://github.com/razee-io/Razeedash-api/tree/master/app/apollo/test should also be updated so that tests with and without the `custom` value are executed to verify both work with the expected data type (see Adam's comment about whether it should be a single string or key value pairs)

app/apollo/schema/channel.js
```diff
@@ -25,6 +25,14 @@ const channelSchema = gql`
     created: Date
     location: String
   }
+  input CustomChannelDataInput {
+    key: String!
+    val: String
+  }
+  type CustomChannelData {
```
michael-topchiev, line 32: Type CustomChannelDataInput is a duplicate of type CustomChannelData, if possible, one of them should be removed.

jscd: From what I understand, I thought you need to separate out GraphQL input types from output types. Is there a different way to handle passing in more complex objects? I've been reading from [here](https://graphql.org/learn/schema/#input-types)

michael-topchiev: i do not know, maybe you can inherit one from another...

jscd: From looking at the [spec](http://spec.graphql.org/June2018/#sec-Input-Objects), it doesn't seem like there's a way for input types to inherit from output types/interfaces or vice-versa. There's ways to extend Input types from other Input types, but that doesn't really help here. I think I could replace the inner `key` and `val` sections using template literals like this:
```
const customChannelType = `
    key: String!
    val: String
`;

const channelSchema = gql`
  // ...
  input CustomChannelDataInput {
    ${customChannelType}
  }
  type CustomChannelData {
    ${customChannelType}
  }
  // ...
`;
```

...but I'm not sure if that would be much cleaner. What do you think?

michael-topchiev: no, let's keep the original. if the types were much more complex then it would make sense.

app/apollo/test/subscriptions.spec.js
```diff
@@ -497,6 +601,57 @@ describe('subscription graphql test suite', () => {
...
+    } catch (error) {
+      if (error.response) {
+        console.error('error encountered:  ', error.response.data);
+      } else {
+        console.error('error encountered:  ', error);
+      }
+      throw error;
+    }
+  });
+
```
michael-topchiev, line 654: All looks good. But if we begin to add three tests for each new GraphQL attribute then test code would be bloated reducing maintainability. Would it be possible to just add the attributes to the existing tests?

app/apollo/schema/channel.js
adamkingit: @rmgraham may have a definitive answer on this. Check with him.

app/apollo/models/channel.schema.js
```diff
@@ -46,6 +46,12 @@ const ChannelSchema = new mongoose.Schema({
     type: Date,
     default: Date.now,
   },
+  custom: [
+    {
+      key: String,
+      val: String,
+    }
+  ],
```
adamkingit, line 54: With apologies, I know I suggested the array, but ....

What happens if a user specifies the same key more than once. Would this work/work better if this was an object?  It would prevent duplicate keys, but with what other implications?

michael-topchiev (commented):
It would be great to not create any redundancy in newly created code.

app/apollo/test/channel.spec.js
```diff
@@ -51,15 +51,46 @@ let presetClusters;
...
+const channel_03_custom = [
+  {
+    'key': 'key_with_null_val',
+    'val': null,
+  },
+  {
+    'key': 'keyB',
+    'val': 'valB',
+  },
+];
```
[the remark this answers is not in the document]
jscd: Oh I see, I didn't notice that was an option. Switching to scalar type JSON would definitely clean this up a ton, I'll start switching over to that.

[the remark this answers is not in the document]
michael-topchiev: Or maybe define new scalar type MAP - see https://www.apollographql.com/docs/apollo-server/schema/custom-scalars/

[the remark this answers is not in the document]
jscd: What's the advantage of using a map over the scalar type JSON?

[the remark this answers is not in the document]
michael-topchiev: Depends on requirements. Maybe you want to ensure users only allowed to pass String values instead of any valid Json. The validations, including duplication of keys, would be done in one place. It's up to you.

michael-topchiev (approved):
looks good to me

jscd:
Travis still isn't triggering, I'm going to try closing and re-opening the PR

jscd:
Replaced with new branch and PR to fix Travis build issues

PR #942, opened 2021-07-26
Allow injecting external auth modules without directly mixing code

carrolp:


carrolp:
As quite a few files touched, here's a helpful summary of the changes:
1. Check and use EXTERNAL_AUTH_MODELS env var in `auth.js`, `user.js`, `index.js`, `organization.js` to load auth packages.  Previously would attempt to load auth packages by assuming the file is in the same directory as built-in auth packages and using a set naming pattern.
2. Change how json files are loaded so that they do not assume they are relative to the current directory (no longer true if `Razeedash-api` is in `node_modules` of a consuming project)
3. `.travis.yml`/`package.json` updated with test for using an external auth module, and associated files in `app/apollo/test/`
4. Some embedded types moved to separate files for clarity (e.g. `resolvers/customScalar.js` was previously inline in `resolvers/index.js`, `schemas/link.js` was previously inline in `schemas/index.js`)
5. Translatable strings json files updated because.... **unclear, I think this was done due to linting execution.  I am revisiting these unintended changes.**

dalehille (approved):

rmgraham (approved):

PR #952, opened 2021-08-10
Kube owner iam lowercase

rmgraham:


dalehille:
could this have an impact on the [clusterAuth->impersonateUser](https://github.com/razee-io/ClusterSubscription/blob/master/lib/remoteResource.js#L58) field from cluster-subscription ?

rmgraham:
Yes, thats the goal. Satellite impersonation is breaking on some users because it only validates properly on lowercase versions of iam usernames. https://github.ibm.com/alchemy-containers/satellite-config/issues/1295

alewitt2 (approved):

PR #955, opened 2021-08-17
enable maintenance mode plugin capability, improve logging

carrolp:


app/utils/maintenance.js
```diff
@@ -16,12 +16,17 @@
 
 const log = require('../log').createLogger('razeedash-api/app/utils/maintenance');
 
-let maintenanceMode = () => false;
+let pluginName = process.env.MAINTENANCE_PLUGIN || './maintenance-default';
+const plugin = require( pluginName );
 
+let maintenanceMode = plugin.maintenanceMode;
+
+// Attempt to load featureflag from legacy mix-in code (this code will be removed once legacy featureflag mix-in code is converted to a plugin)
```
topchiev-ibm, line 24: Rather than creating new change and test it later, why not remove legacy code right now to make the solution final?

carrolp: > why not remove legacy code right now 

So that the change doesn't introduce breakage/change if Razeedash-api is merged but Satellite-Config is not.

topchiev-ibm: > 
> 
> > why not remove legacy code right now
> 
> So that the change doesn't introduce breakage/change if Razeedash-api is merged but Satellite-Config is not.

Old version of SatCon refers to old version of API and will not be broken, new version of SatCon will refer to this version and will not be broken either. All should be good.

carrolp: We **are** going to be updating satcon to adopt this plugin model and picking up this new razeedash-api, so accepting this suggestion.  Risk is low that satcon would need an updated razeedash-api before those changes are merged in satcon.

adamkingit (approved):

PR #956, opened 2021-08-19
Changes to S3 client for high-performance resource cleanup

michael-topchiev:


michael-topchiev:
This change is required for resource history cleanup job to work when it is enabled.

app/storage/s3NewClientFactory.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright 2019 IBM Corp. All Rights Reserved.
```
adamkingit, line 2:
```suggestion
 * Copyright 2021 IBM Corp. All Rights Reserved.
```

app/storage/s3NewClient.js
```diff
@@ -80,6 +80,19 @@ module.exports = class S3NewClient {
     }).promise();
   }
 
+  async deleteMultipleObjects(bucketName, keys) { // array of string keys
```
adamkingit, line 83:
```suggestion
  async deleteObjects(bucketName, keys) { // array of string keys
```

app/storage/s3NewClient.js
```diff
@@ -80,6 +80,19 @@ module.exports = class S3NewClient {
...
+    const objects = [];
+    for (const k of keys) {
+      objects.push({ Key: k });
+    }
```
adamkingit, line 87: const objects = keys.map(e => { return {Key: e}; } );

app/storage/s3NewClient.js
adamkingit: Do we need any type checking to ensure that keys is an Array?

adamkingit (changes requested):
Questions and suggestions added to individual lines

app/storage/s3NewClient.js
michael-topchiev: the reason this function and all other functions in this file do not validate input parameters is because they all have only one client, used only internally and do not accept unknown input.

app/storage/s3NewClientFactory.js
```diff
@@ -0,0 +1,43 @@
...
+'use strict';
+
+const conf = require('./../conf.js').conf;
+const S3ClientClass = require('./s3NewClient');
+
+module.exports = (logger, location, endpoint) => { // only logger parameter is mandatory
+  location = location || conf.storage.defaultLocation;
+  if (location) {
+    location = location.toLowerCase();
+  }
```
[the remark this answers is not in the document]
michael-topchiev: added error message

carrolp (approved):

adamkingit (approved):

PR #972, opened 2021-09-15
Encryption buckets

rmgraham:


app/apollo/resolvers/channel.js
```diff
@@ -309,8 +309,13 @@ const channelResolvers = {
...
       const path = `${org_id.toLowerCase()}-${channel.uuid}-${name}`;
-      const bucketName = conf.storage.getChannelBucket(channel.data_location);
-      const handler = storageFactory(logger).newResourceHandler(path, bucketName, channel.data_location);
+      const { data_location=null } = channel;
+      const bucketConfObj = {
+        type: 'active',
+        region: data_location,
+        kind: 'configs',
+      };
+      const handler = storageFactory(logger).newResourceHandler({ path, bucketConfObj, data_location });
```
carrolp, line 318: `data_location` is part of the `bucketConfObj` _and_ a separate arg to `newResourcehandler`.  Should eliminate from one of the two... preferably keep it in the conf obj only unless there's a technical reason it can't be removed from the function signature.
```suggestion
      const handler = storageFactory(logger).newResourceHandler({ path, bucketConfObj });
```

app/conf.js
```diff
@@ -42,7 +42,23 @@ const conf = {
   maintenance: {
     flag: process.env.MAINTENANCE_FLAG,
     key: process.env.MAINTENANCE_KEY
-  }
+  },
+  kms: {
+    server: {
+      serviceUrl: process.env.KMS_SERVER_SERVICE_URL || 'https://us-south.kms.cloud.ibm.com',
+      iamAuthUrl: process.env.KMS_SERVER_IAM_AUTH_URL || 'https://iam.cloud.ibm.com',
```
carrolp, line 49: Can we avoid adding IBM cloud defaults, especially `us-south` specific defaults?  If possible, avoid needing a default at all -- if the value isn't set, code shouldn't be attempting to use KMS.
```suggestion
      serviceUrl: process.env.KMS_SERVER_SERVICE_URL,
      iamAuthUrl: process.env.KMS_SERVER_IAM_AUTH_URL,
```

app/conf.js
```diff
@@ -42,7 +42,23 @@ const conf = {
...
+      apiKey: process.env.KMS_SERVER_API_KEY,
+      bluemixInstanceGuid: process.env.KMS_SERVER_BLUEMIX_INSTANCE_GUID,
+      rootKeyId: process.env.KMS_SERVER_ROOT_KEY_ID,
+    },
+    cos:{
+      enabled: process.env.KMS_COS_ENABLED,
```
carrolp, line 55: Using 'enabled' env vars as booleans always makes me nervous as it's likely someone could try to `export KMS_COS_ENABLED=false` and it would still treat it as true.
```suggestion
      enabled: process.env.KMS_COS_ENABLED && !(process.env.KMS_COS_ENABLED.trim().toLowerCase() === 'false'),
```

app/routes/v2/clusters.js
```diff
@@ -122,13 +122,25 @@ var runAddClusterWebhook = async(req, orgId, clusterId, clusterName)=>{
...
 
-function pushToS3Sync(key, searchableDataHash, dataStr, data_location, logger) {
+function pushToS3Sync(org, key, searchableDataHash, dataStr, data_location, logger) {
   //if its a new or changed resource, write the data out to an S3 object
   const result = {};
-  const bucket = conf.storage.getResourceBucket(data_location);
+  // const bucket = conf.storage.getResourceBucket(data_location);
+  const locationConf = conf.storage.s3ConnectionMap.get(data_location || conf.storage.defaultLocation) || {};
+  const bucketPrefix = locationConf.orgBucketPrefix || 'razee-org-';
+  const bucket = `${bucketPrefix}${org._id.toLowerCase()}`;
```
carrolp, line 131: Why toLowerCase?  I wouldn't think uppercase letters would be a problem for bucket names, and it unnecessarily obscures the 'true' org ID.  I'm not aware of any code dependent on parsing the 'true' org ID with correct capitalization from the name (the COS backup code no longer does for example), but this could introduce unnecessary confusion for SREs.
```suggestion
  const bucket = `${bucketPrefix}${org._id}`;
```

app/routes/v2/clusters.js
```diff
@@ -122,13 +122,25 @@ var runAddClusterWebhook = async(req, orgId, clusterId, clusterName)=>{
...
+    region: data_location,
+    kind: 'resources',
+  };
   const hash = crypto.createHash('sha256');
   const keyHash = hash.update(JSON.stringify(key)).digest('hex');
-  const handler = storageFactory(logger).newResourceHandler(`${keyHash}/${searchableDataHash}`, bucket, data_location);
+  const path = `${org._id}/${keyHash}/${searchableDataHash}`;
+  console.log(22222, data_location, locationConf, org);
+  const factory = storageFactory({ logger });
+  const handler = factory.newResourceHandler({ path, bucketConfObj, data_location, org });
```
carrolp, line 143: Same comment as earlier: avoid duplicate data_location in both the conf obj and the new handler function signature.

app/storage/s3ResourceHandler.js
```diff
@@ -19,91 +19,186 @@
...
 
 class S3ResourceHandler {
+  logger = null;
 
-  constructor(logger, resourceKey, bucketName, location, endpoint) {
-    this.logger = logger;
-    if (!resourceKey || !bucketName) {
-      throw new Error(`Path (${resourceKey}) and/or bucket name (${bucketName}) is not specified`);
+  constructor(args){
+    var { logger, path, bucketConfObj, endpoint, org } = args;
```
carrolp, line 30: Use `const` instead of `var` in this and many other locations in the PR
```suggestion
    const { logger, path, bucketConfObj, endpoint, org } = args;
```

app/storage/s3ResourceHandler.js
```diff
@@ -19,91 +19,186 @@
...
+    else{
+      throw new Error(`unhandled type "${type}"`);
+    }
+    await models.Organization.updateOne({ _id: org._id }, { $set: sets });
+    return bucketKey;
   }
 
   async setDataAndEncrypt(stringOrBuffer, key) {
-    this.logInfo(`Uploading object ${this.bucketName}:${this.resourceKey} ...`);
+    var { path, org } = this;
```
carrolp: Additional examples where `const` would be preferred over `var`

app/storage/storageConfig.js
```diff
@@ -56,17 +69,19 @@ class StorageConfig {
     }
   }
 
-  getChannelBucket(location) {
+  getChannelBucket({ location }) {
+    throw `can we deprecate this`
```
carrolp, line 73: I agree that the gist of this PR means this can/should be removed entirely.

app/storage/storageFactory.js
```diff
@@ -64,8 +64,17 @@ class StorageFactory {
...
+  newResourceHandler(args={}) {
+    var { path, bucketConfObj, location, endpoint, org } = args;
+    // if(!path || !bucketConfObj){
+    //   throw new Error(`required args: { path, bucketConfObj }`);
+    // }
+    console.log(44444, { path, bucketConfObj, location, endpoint, org })
+    // args is an obj of other attrs to pass that can be useful in certain contexts, such as { org }
+    const targetHandler = config.defaultHandler.constructor({
+      ..._.pick(this, ['logger']),
+      ...args
```
carrolp, line 76: This is interesting syntax, but (to me at least) not intuitive.  A code comment explaining what it's doing, for newbs like me, would be appreciated.

app/utils/orgs.js
```diff
@@ -68,4 +71,55 @@ const decryptOrgData = (orgKey, data) => {
...
+  var updates = {
+    $set: {
+      'kms.crn': crn,
+    }
+  };
+  await models.Organization.updateOne(search, updates);
+  return crn;
+};
+
+var rotateKmsKeyForOrg = async({ org, metroConf })=>{
```
carrolp, line 100: When is `rotateKmsKeyForOrg` called?  I suspect we need a 'featurespec' describing the expected flow (e.g. I'm used to keys being generated by SREs and stored elsewhere rather than auto-generated and presumably auto-rotated... and not clear where the rotation would occur when there are multiple instances accessing the buckets, etc).

app/storage/storageConfig.js
```diff
@@ -17,12 +17,17 @@
 
 'use strict';
 
+var _ = require('lodash');
```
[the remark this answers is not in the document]
carrolp: Still var, should be const?

app/storage/storageConfig.js
```diff
@@ -56,17 +68,11 @@ class StorageConfig {
     }
   }
 
-  getChannelBucket(location) {
-    location = location ? location.toLowerCase() : this.defaultLocation;
-    const connection = this.s3ConnectionMap.get(location);
-    return connection ? connection.channelBucket : undefined;
-  }
-
   getResourceBucket(location) {
```
carrolp, line 71: Can `getResourceBucket` be removed just like `getChannelBucket`?

app/storage/storageConfig.js
```diff
@@ -56,17 +68,11 @@ class StorageConfig {
...
     location = location ? location.toLowerCase() : this.defaultLocation;
     const connection = this.s3ConnectionMap.get(location);
     return connection ? connection.resourceBucket : undefined;
   }
```
carrolp, line 75: Can `getResourceBucket` be removed just like `getChannelBucket`?

app/storage/storageConfig.js
```diff
@@ -36,8 +41,15 @@ class StorageConfig {
         connection.locationConstraint = env['S3_' + metro + '_LOCATION_CONSTRAINT'];
         connection.s3ForcePathStyle = true;
         connection.signatureVersion = 'v4';
-        connection.channelBucket = env['S3_' + metro + '_CHANNEL_BUCKET'] || env.S3_CHANNEL_BUCKET || 'razee';
         connection.resourceBucket = env['S3_' + metro + '_RESOURCE_BUCKET'] || env.S3_RESOURCE_BUCKET || connection.channelBucket || 'razee';
+        connection.orgBucketPrefix = env['S3_' + metro + '_ORG_BUCKET_PREFIX'] || env.S3_ORG_BUCKET_PREFIX || 'razee-org-';
+        let kmsEnabled = (env['S3_' + metro + '_KMS_ENABLED'] || env.S3_KMS_ENABLED || null);
+        kmsEnabled = !!(kmsEnabled && !_.includes(['false', '0'], kmsEnabled));
```
carrolp, line 47: app/conf.js is using `process.env.KMS_COS_ENABLED`, here it's checking both `S3_[metro]_KMS_ENABLED` and `S3_KMS_ENABLED` -- there should be only one env var for whether KMS is enabled.

app/conf.js is also using slightly simpler code to check if the ENABLED value is true-ey or false-ey, that doesn't require using `let` or lodash.  Should be the same in both places (and simpler version preferred personally, so I don't have to do so much mental counting of `!`s to understand what's going on :-D )

app/storage/storageConfig.js
rmgraham: I had removed it. But the old s3 code uses that one still, so I had to add it back.

app/storage/s3ResourceHandler.js
```diff
@@ -19,91 +19,113 @@
...
+    if(!bucketConfObj && !bucketName){
+      throw new Error('Pass a bucketConfObj or bucketName');
+    }
+    bucketConfObj = bucketConfObj || {};
+    if(location){
+      bucketConfObj.location = location;
+    }
+    if(!bucketName){
+      bucketName = conf.storage.getResourceBucket(location);
+    }
```
carrolp, line 37:
```suggestion
```

app/utils/orgs.js
```diff
@@ -68,4 +70,55 @@ const decryptOrgData = (orgKey, data) => {
...
+  return true;
+};
+
+// setTimeout(async()=>{
+//   var org = await models.Organization.findOne({ _id: 'k4TTY77XNPMGjppfW' });
+//   var result = await rotateKmsKeyForOrg({ org });
+//   console.log(77777, org, result)
+// },1)
+
+module.exports = { getOrg, verifyAdminOrgKey, encryptOrgData, decryptOrgData, getKmsKeyForOrg, rotateKmsKeyForOrg };
```
carrolp, line 124:
```suggestion

module.exports = { getOrg, verifyAdminOrgKey, encryptOrgData, decryptOrgData, getKmsKeyForOrg };
```

app/storage/storageConfig.js
carrolp: See proposed change below to s3ResourceHandler.js so that we can remove getResourceBucket entirely.  It shouldn't be used once we change to have a separate bucket per org.

app/utils/orgs.js
carrolp: Suggest removing this unused code.  Instead I suspect it would be better to configure Key Protect to automatically rotate the root key monthly... after consulting with others with more expertise.

app/storage/s3ResourceHandler.js
rmgraham: We support multiple different storage handlers. That function's for those using `s3`. The new one that this PR creates is `s3OrgBucket`. I had to create the new one so the old one can continue to handle existing db data properly.

app/utils/orgs.js
rmgraham: Lets put it in just incase we need it. If we delete it now, it might get squashed and we'll lose the entire function and have to rewrite it later. idc if you merge and then immediately delete it in a new pr.

app/storage/s3ResourceHandler.js
```diff
@@ -19,91 +19,113 @@
...
+const _ = require('lodash');
 
 class S3ResourceHandler {
 
-  constructor(logger, resourceKey, bucketName, location, endpoint) {
+  constructor(args) {
+    let { logger, path, bucketName=null, bucketConfObj=null, location, endpoint, org } = args;
+    if(!bucketConfObj && !bucketName){
+      throw new Error('Pass a bucketConfObj or bucketName');
+    }
```
carrolp, line 30: GIT's diff is a bit garbled, but the original code was:
```
if (!resourceKey || !bucketName) {
      throw new Error(`Path (${resourceKey}) and/or bucket name (${bucketName}) is not specified`);
}
```
By changing it to `&&` it changes the behavior and makes it confusing.  E.g. why would it be necessary to construct an s3ResourceHandler without specifying the bucketName and have it default to using the common (not per-org, not encrypted) `resources` bucket?  That wasn't possible before.

carrolp (approved):

carrolp (dismissed):

carrolp (approved):

PR #1005, opened 2021-10-15
Fix security issues and change swagger.json to be compatible with owasp zap tool

michael-topchiev:


carrolp:
Both the before and after appear to be valid swagger -- is there some documentation or similar that shows why owasp zap test tools cannot handle the original valid swagger using `basePath`?

michael-topchiev:
> 
> 
> Both the before and after appear to be valid swagger -- is there some documentation or similar that shows why owasp zap test tools cannot handle the original valid swagger using `basePath`?

There is likely a bug in the tool, but we spent too much time dealing with the problem already, no time to wait their fix.

adamkingit (approved):

app/routes/index.js
```diff
@@ -81,7 +81,7 @@ router.use(async (req, res, next) => {
   if(!orgKey){
     orgKey = req.query.orgKey;
     if(!orgKey){
-      return res.status(401).send( 'razee-org-key required' );
+      return res.status(401).type('json').send( 'razee-org-key required' );
```
adamkingit, line 84:
```suggestion
      return res.status(401).type('json').send( {msg: "razee-org-key required"} );
```

PR #1040, opened 2021-11-18
Rest -> graphql

rgnova:
Adds rest apis for some of the more desired graphql apis.
Requests use the expressjs `app.handle()` feature to proxy through from a regular rest api path to the graphql app.

### DIFF
--- a/app/routes/v1/gql.js
+++ b/app/routes/v1/gql.js
@@ -1,0 +1,247 @@
+const express = require('express');
+const router = express.Router();
+const asyncHandler = require('express-async-handler');
+const mongoConf = require('../../conf.js').conf;
+const MongoClientClass = require('../../mongo/mongoClient.js');
+const MongoClient = new MongoClientClass(mongoConf);
+const storageFactory = require('./../../storage/storageFactory');
+const getOrg = require('../../utils/orgs.js').getOrg;
+const _ = require('lodash');
+const mainServer = require('../../');
+
+
+var sendReqToGraphql = async({ req, res, query, variables, operationName })=>{
+  req.path = req.url = req.originalUrl = '/graphql';
+  req.body = {
+    query,
+    variables,
+    operationName,
+  };
+  req.method = 'POST';
+  console.log(555555, req.path, req.body, req.headers)
+  mainServer.app.handle(req, res, {});
+};
+
+var getOrgId = (req, res, next)=>{
+  const orgId = req.get('org-id') || req.body.orgId || req.query.orgId;
+  if(!orgId){
+    throw new Error(`Please pass an orgId in an "org-id" header, an "orgId" post body param, or an orgId query string attribute`);
+  }
+  req.orgId = orgId;
+  next();
+};
+
+
+router.post('/channels', getOrgId, asyncHandler(async(req, res)=>{
+  // #swagger.tags = ['channels']
+  // #swagger.summary = 'Adds a new channel'
+  var { orgId } = req;
+  var operationName  = 'addChannel';
+  var query = `
+    mutation ${operationName}($orgId:  String!, $name: String!) {
+      addChannel(orgId: $orgId, name: $name) {
+        uuid
+      }
+    }
+  `;
+  var name = req.body.name;
+  if(!name){
+    throw new Error(`needs { name }`);
+  }
+  var variables = {
+    orgId,
+    name,
+  };
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
+
+router.get('/channels', getOrgId, asyncHandler(async(req, res)=>{
+  // #swagger.tags = ['channels']
+  // #swagger.summary = 'Gets all channels for an org'
+  var { orgId } = req;
+  var operationName  = 'channels';
+  var query = `
+    query ${operationName}($orgId: String!) {
+      channels(orgId: $orgId) {
+        uuid
+        orgId
+        name
+        created
+        versions {
+          name
+          description
+          uuid
+          created
+        }
+      }
+    }
+
+  `;
+  var variables = {
+    orgId,
+  };
+  console.log(3333, query, variables, operationName)
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
+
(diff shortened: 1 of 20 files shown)


app/routes/v1/gql.js
```diff
@@ -0,0 +1,239 @@
...
+  if(!uuid || !clusters){
+    throw new Error('needs { uuid, clusters }');
+  }
+  const variables = {
+    orgId,
+    uuid,
+    clusters,
+  };
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
```
carrolp, line 178: `/groups/groupClusters` is not proper syntax for a REST request.  Instead, client should be able to:
- POST `/groups` to create a group, specifying the name _and the clusters_ in the body
- PUT `/groups/[uuid]` to update a group, specifying the new clusters array in the body

If the graphql API doesnt allow creating a group with the clusters list, and the clusters list must be updated separately, it may even be necessary to make _two_ graphql calls for a single `POST`.

app/routes/v1/gql.js
```diff
@@ -0,0 +1,239 @@
...
+const sendReqToGraphql = async({ req, res, query, variables, operationName })=>{
+  req.path = req.url = req.originalUrl = '/graphql';
+  req.body = {
+    query,
+    variables,
+    operationName,
+  };
+  req.method = 'POST';
+  mainServer.app.handle(req, res, {});
+};
```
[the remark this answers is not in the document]
carrolp: Note especially the code comment:
```
// INSUFFICIENT, should return 201 on creation, along with a `Location` header.
```
A complete REST interface should not have a POST implementation without a GET implementation
- E.g.: Successful POST to create on `/channels` should return a response with 201 and a `Location: /channels/[uuid]` header.  The client should be able to GET that `/channels/[uuid]` and retrieve the channel.

The function to send the graphql request needs to be aware of whether it's doing a 'create' or a 'query' and be able to adjust its return code appropriately based on the graphql response.

Reference: https://restfulapi.net/http-methods/#summary

_Initially_ we could skip the `Location` header **if there is no GET implemented yet**, but must ensure the correct response codes are used.
  - Note that while PR doesn't contain GET for all POSTs, it does for _some_.  `subscriptions` has both a POST and a GET -- the POST should return the 201 and `Location` header if successful.

app/swagger.json
```diff
@@ -622,6 +622,235 @@
...
+                  "example": "any"
+                }
+              }
+            }
+          }
+        ],
+        "responses": {}
+      }
+    },
+    "/api/v1/clusters/clustersByOrgId": {
```
carrolp, line 711: Update to match proposed changes to REST interface in gql.js

app/swagger.json
```diff
@@ -622,6 +622,235 @@
...
+        "parameters": [
+          {
+            "name": "obj",
+            "in": "body",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "name": {
+                  "example": "any"
+                }
```
carrolp, line 751: Should be able to POST clusterIDs list as part of creation, ref gql.js proposed changes.

app/swagger.json
```diff
@@ -622,6 +622,235 @@
...
+              }
+            }
+          }
+        ],
+        "responses": {}
+      }
+    },
+    "/api/v1/groups/groupClusters": {
```
carrolp, line 759: Ensure updated to match gql.js proposed changes.

app/routes/v1/gql.js
```diff
@@ -0,0 +1,239 @@
...
+    throw new Error('needs { name }');
+  }
+  const variables = {
+    orgId,
+    name,
+  };
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
+
+router.get('/channels', getOrgId, asyncHandler(async(req, res)=>{
```
carrolp, line 51: Ideally should implement `/channels/:uuid` and not _just_ 'get all'

app/routes/v1/gql.js
```diff
@@ -0,0 +1,239 @@
...
+    name,
+    groups,
+    clusterId,
+    channelUuid,
+    versionUuid,
+  };
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
+
+router.get('/subscriptions/:uuid', getOrgId, asyncHandler(async(req, res)=>{
```
carrolp, line 211: Ideally should implement `/subscriptions` (get all) not just 'get one'

carrolp (dismissed):

app/routes/v1/gql.js
```diff
@@ -0,0 +1,239 @@
...
+    orgId,
+    channelUuid,
+    name,
+    type,
+    content,
+  };
+  sendReqToGraphql({ req, res, query, variables, operationName });
+}));
+
+router.get('/clusters/clustersByOrgId', getOrgId, asyncHandler(async(req, res)=>{
```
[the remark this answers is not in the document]
carrolp: Fixed by latest commits

app/routes/v1/gql.js
carrolp: Fixed by latest commits, other than allowing specifying clusters in the POST to `/groups`.  That will require more work, perhaps ensuring multiple graphql queries can be sent without returning a response to the client (one to create the group, one to set the clusters, and only sending back 201 CREATED response if both succeeded, else 400 or similar).

At this time I would propose including detailed code comments at least explaining that only the name can be specified on the POST to /groups and why it cannot include clusters.

app/routes/v1/gql.js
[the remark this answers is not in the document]
carrolp: Resolved by latest commits

app/swagger.json
carrolp: Not at this time

app/routes/v1/gql.js
carrolp: latest commit added code comment explaining current limitation.

carrolp (approved):
Approving, but we should get all three reviews approved before squash-merging

rgnova:
I cant approve since its my PR. But looks good 👍

adamkingit (approved):

PR #1041, opened 2021-11-19
extend graphql serviceSubscription query to include cluster details

carrolp:
https://github.ibm.com/alchemy-containers/satellite-config/issues/1234

### DIFF
--- a/app/apollo/models/cluster.schema.js
+++ b/app/apollo/models/cluster.schema.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+const _ = require('lodash');
+
 const mongoose = require('mongoose');
 const { CLUSTER_REG_STATES } = require('./const');
 const ClusterSchema = new mongoose.Schema({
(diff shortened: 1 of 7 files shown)


adamkingit (approved):

app/apollo/resolvers/serviceSubscription.js
```diff
@@ -96,6 +96,16 @@ const serviceResolvers = {
...
+      // render cluster information if users ask for
+      if(queryFields.cluster && serviceSubscriptions) {
+        const clusterIds = _.map(serviceSubscriptions, 'clusterId');
+        const clusters = await models.Cluster.getClustersByIds(clusterIds);
+        serviceSubscriptions = serviceSubscriptions.map((sub)=>{
+          sub.cluster = clusters[sub.clusterId];
+          return sub;
+        });
+      }
+
```
rgnova, line 108: This is best off inside of the applyQueryFieldsToSubscriptions() function
The applyQueryFields functions call eachother when necessary, which lets you do deeper nesting of requested attributes. So if a user was accessing a different resolver which returned serviceSubscriptions, they can also ask for cluster info inside of those, and it will know to run this block of code to include them.

app/apollo/models/cluster.schema.js
```diff
@@ -111,4 +113,24 @@ const ClusterSchema = new mongoose.Schema({
...
+  const clusters = await this.find({ cluster_id: { $in: ids } }, { }, { lean: 1 });
+  const mappedClusters = clusters.map((cluster)=>{
+    //return a BasicCluster
+    return {
+      id: cluster._id,
+      orgId: cluster.org_id,
+      clusterId: cluster.cluster_id,
+      name: cluster.name,
+      registration: cluster.registration,
+    };
```
rgnova, line 130: Is there a reason to filter to basic clusters? Maybe for security?
The typical graphql pattern sends back the full object, and the user must explicitly request every attribute they want. So filtering like this doesnt make things easier (like they usually would in rest)

rgnova (changes requested):

app/apollo/models/cluster.schema.js
carrolp: Yes, security was my intent.  The 'full' Cluster type includes `resources`, which are customer data that doesn't need to be exposed when querying serviceSubscriptions.  Other Cluster type attributes such as `status`, and `groups` are not necessary for the serviceSubscriptions query, so also excluded.

app/apollo/resolvers/serviceSubscription.js
carrolp: Done

adamkingit (approved):

rgnova (approved):

PR #1050, opened 2021-12-03
put updateClusterResources on graphql/mongoose. includes a /v3/ PUT rest endpoint

rgnova:
- adds `updateClusterResources(orgId: String! @sv, clusterId: String! @sv, resourceChanges: [ResourceChange]!): UpdateClusterResourcesResponse`
- adds `PUT /v3/clusters/:clusterId/resources` as well, which proxies to the graphql

These implement the v2 version of `updateClusterResources()`. But now we use mongoose, which means we can now enforce mongoose schemas.


app/apollo/models/resource.schema.js
```diff
@@ -43,13 +43,20 @@ const ResourceSchema = new mongoose.Schema({
     default: false,
   },
   searchableData: {
-    type: Map,
+    type: Object,
```
carrolp, line 46: What's the implication of changing Map to Object?  This seems like it should be extraneous to adding `/v3/:cluster_id/resources` and adding to GQL.

app/apollo/models/resource.schema.js
```diff
@@ -43,13 +43,20 @@ const ResourceSchema = new mongoose.Schema({
...
     type: String,
     default: '',
   },
+  searchableDataHist: {
+    type: Object,
+    default: {},
+  },
+  histId: {
+    type: String,
+  },
```
carrolp, line 59: Why are these additional database fields needed now?  They weren't used previously (obviously, since they didn't exist) and I do not see any changes using them in this PR either.

app/apollo/resolvers/resource.updateClusterResources.js
```diff
@@ -0,0 +1,296 @@
...
+                  }
+                  changes = {
+                    $set: { deleted: false, hash: resourceHash, histId, data: dataStr, searchableData: searchableDataObj, searchableDataHash: searchableDataHash },
+                    $currentDate: { created: true, updated: true, lastModified: true },
+                    ...pushCmd
+                  };
+                  options = { upsert: true };
+                  start = Date.now();
+                  // we need to use table.collection.updateOne(), because mongoose doesnt let you upsert when you dont set the primary key
+                  models.ResourceStat.collection.updateOne({ org_id: org._id }, { $inc: { deploymentCount: 1 } }, { upsert: true });
```
carrolp, line 200: When
- GQL code processes mutation `updateClusterResources` and
- a resource is POLLED, MODIFIED, or ADDED, and
- the resource is not already in the database,

the `deploymentCount` counter for the Org is incremented by one.

The ResourceStats related code should be removed.
- The statistic isn't _useful_
  - Nothing ever _decrements_ the `deploymentCount` counter so it doesn't represent how many resources are currently in an org.
  - The total number of resources ever added in an org, at any point in history, for any cluster, even if resource and cluster subsequently deleted, doesn't provide useful information.
- Nothing _allows_ querying the `ResourceStats` for an org (no REST or GQL endpoints).
- Nothing _requires_ querying the `ResourceStats` for an org (not show in the CLI or GUI, not logged).
- `ResourceStatSchema` includes `clusterCount` and `deployablesCount` attributes that are unused.

app/routes/v2/clusters.js
```diff
@@ -183,6 +183,8 @@ const syncClusterResources = async(req, res)=>{
   res.status(200).send('Thanks');
 };
 
+
+// todo: deprecate this after forcing updater to use the graphql/v3 endpoint
```
carrolp, line 187: The overall intent is not clear.  The implementation implication is to replace `POST /v2/:cluster_id/resources` with `PUT /v3/:cluster_id/resources`.  However:
- `/v2` APIs use org-key authentication while GQL and `/v3` APIs use user bearer token IIRC.
- There isn't a plan to deprecate `/v2` APIs and update all `watch-keeper` instances to use the `/v3` replacement and different auth.

Wouldn't it be better to simply update the existing `/v2` cluster resources implementation to call GQL instead?

app/routes/v3/gql.js
```diff
@@ -532,4 +532,35 @@ const getSubscription = async (req, res) => {
 };
 router.get('/subscriptions/:uuid', getOrgId, asyncHandler(getSubscription));
 
+
+const updateClusterResources = async (req, res) => {
```
carrolp, line 536: Automated tests and swagger updates needed.

app/apollo/resolvers/resource.updateClusterResources.js
```diff
@@ -0,0 +1,296 @@
+/**
+ * Copyright 2020 IBM Corp. All Rights Reserved.
```
carrolp, line 2: New files should have a copyright of 2021 rather than 2020.

_Existing_ files that are updated also need copyright statements updated appropriately.

app/apollo/resolvers/index.js
```diff
@@ -22,6 +22,7 @@ This allows, for example, replacing the built in 'user' resolver with a custom o
 const resolversMap = {
   'user': './user',
   'resource': './resource',
+  'resource.updateClusterResources': './resource.updateClusterResources',
```
carrolp, line 25: I think it would make more sense for the cluster resources handling to be part of the existing `./cluster` rather than a new file with an unusual naming pattern.  Or perhaps even better, as a new `Mutation` section in `./resource.js`, where `Query` and `Subscription` on resources is already defined.  Is there a reason to keep it separate and so 'specialized', e.g. so it can be overridden in some scenario?

carrolp (changes requested):
Multiple inline questions (e.g. what's the big picture for this change) and requested changes (e.g. needs copyright updates, automated tests, swagger updates).

PR #1074, opened 2022-01-27
allow null cluster on serviceSubscription query,

carrolp:
e.g. for if the cluster was deleted but the service subscription still references it


ethanstjohn (approved):
Looks good to me!

PR #1175, opened 2022-07-08
more strictly enforce special characters in identifiers

carrolp:
https://github.ibm.com/alchemy-containers/satellite-config/issues/1587


carrolp:
Closing PR -- the Razeedash-api test cases are riddled with test cases using 'invalid' names e.g. https://github.com/razee-io/Razeedash-api/blob/master/app/apollo/test/channel.spec.js#L424.  Not comfortable making a breaking change here for little benefit.

PR #1191, opened 2022-08-09
log db close on shutdown

carrolp:
PR unit tests sometimes failing with errors like:
```
  1) subscription graphql test suite
       "after all" hook for "remove a subscription":
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/alchemy-containers/satellite-config/node_modules/razeedash-api/app/apollo/test/subscriptions.spec.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
  2) user graphql
       "before all" hook in "user graphql":
     Uncaught Error: listen EADDRINUSE: address already in use :::18006
      at Server.setupListenHandle [as _listen2] (node:net:1372:16)
```
I.e. the first test fails to stop the server within 50 s and then the next test fails to start the server.  Exact cause is unclear, but is likely related to the update from mongoose v5.x to v6.x.  Adding additional logging... **and taking extra steps to close and disconnection from the database on shutdown**.

### DIFF
--- a/app/apollo/index.js
+++ b/app/apollo/index.js
@@ -211,8 +211,11 @@ const createSubscriptionServer = (httpServer, apolloServer, schema) => {
 
 
 const stop = async (apollo) => {
-  await apollo.db.connection.close();
-  await apollo.server.stop();
+  console.log('🏄 Closing database connection');
+  await apollo.db.connection.close( true ); // Use 'true' to force close and prevent timeout during unit tests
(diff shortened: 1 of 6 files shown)


ethanstjohn (approved):
Very thorough!

PR #1234, opened 2022-10-19
system subscription for operators

kdai7:


package-lock.json
```diff
@@ -1,8 +1,16958 @@
 {
   "name": "razeedash-api",
   "version": "0.6.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
carrolp, line 4: The lockfileVersion changing from 1 to 2 suggests you updated locally while using Node 16 rather than Node 14, which is still specified in the `.travis.yml` file.  Moving from 14 to 16 might be a good idea, but should be separate from this PR (might have unintended consequences).

app/routes/v1/systemSubscriptions.js
```diff
@@ -39,8 +41,77 @@ type: Opaque
...
+  const watchKeeperYaml = await request(url);
+
+  res.status( 200 ).send( watchKeeperYaml );
+};
+
 // /api/v2/systemSubscriptions/primaryOrgKey
 router.get('/primaryOrgKey', getOrg, asyncHandler(getPrimaryOrgKeySubscription));
+router.get('/clusterSubscription', asyncHandler(getClusterSubscriptionSubscription));
+router.get('/remoteResource', asyncHandler(getRemoteResourceSubscription));
+router.get('/watchKeeper', asyncHandler(getWatchKeeperSubscription));
```
carrolp, line 114: Would it be possible to have a single System Subscription to upload all the operators together?  I don't think we have a need to separately update cs, rr, wk.... but it may have some benefit I don't see yet?

carrolp (approved):

PR #1279, opened 2023-01-10
eliminate all use of channel.versions attribute

carrolp:


app/apollo/resolvers/channel.js
```diff
@@ -211,15 +207,14 @@ const channelResolvers = {
       await validAuth(me, org_id, ACTIONS.CREATE, TYPES.CHANNEL, queryName, context);
 
       // Create the channel object to be saved.
-      const uuid = UUID();
       const kubeOwnerId = await models.User.getKubeOwnerId(context);
       const newChannelObj = {
         _id: UUID(),
-        uuid,
+        uuid: UUID(),
```
adamkingit, line 213: Do we want  `_id != uuid` ?

carrolp: It could, though this would be a change from previous behavior.  The separate values for `uuid` and `_id` _are_ confusing, but making them use the same value might make future editors believe they can switch between using the two for queries for example, which would cause problems with old data created when they were different.

adamkingit: I see they are diff for other objects (in other code), but this code previously explicitly kept them the same.

app/apollo/resolvers/channel.js
```diff
@@ -330,37 +325,27 @@ const channelResolvers = {
...
         // Attempt to create subscription(s)
         await Promise.all( subscriptions.map( async (s) => {
           const version = versions.find( v => v.name === s.versionName );
+          if( !version ) {
+            logger.error({ req_id, user: whoIs(me), org_id, name }, `${queryName} unable to create subscription '${s.name}' when serving ${req_id}, version '${s.versionName}' was not created.`);
+          }
+          else if( !version.uuid ) {
+            logger.error({ req_id, user: whoIs(me), org_id, name }, `${queryName} unable to create subscription '${s.name}' when serving ${req_id}, version '${s.versionName}' failed creation.`);
+          }
+
```
adamkingit, line 348:
```suggestion
          if( !version || !version.uuid) {
            logger.error({ req_id, user: whoIs(me), org_id, name }, `${queryName} unable to create subscription '${s.name}' when serving ${req_id}, version '${s.versionName}' was not created.`);
          }
```

app/apollo/resolvers/channel.js
```diff
@@ -390,13 +375,13 @@ const channelResolvers = {
             subscriptionsRbacSync( [subscriptionObj], { resync: false }, context ).catch(function(){/*ignore*/});
           }
           catch( e ) {
-            logger.error(e, `${queryName} failed to create subscription '${subscriptionObj.name}' when serving ${req_id}.`);
-            // Cannot fail here, the Version has already been created.  Continue.
+            logger.error(e, `${queryName} error creating subscription '${subscriptionObj.name}' when serving ${req_id}.`);
+            // Cannot fail here, the Channel has already been created.  Continue.
```
adamkingit, line 379: Do we want to log an error here if this is an expected, continuable "exception?"

app/apollo/resolvers/channel.js
```diff
@@ -604,7 +573,7 @@ const channelResolvers = {
           subscriptionsRbacSync( [subscription], { resync: false }, context ).catch(function(){/*ignore*/});
         }
         catch( e ) {
-          logger.error(e, `${queryName} failed to create subscription '${subscription.name}' when serving ${req_id}.`);
+          logger.error(e, `${queryName} error creating subscription '${subscription.name}' when serving ${req_id}.`);
```
adamkingit, line 576: Do we want to log an error here if this is an expected, continuable "exception?" Should be no worse than a warn, if logged at all.

app/apollo/utils/versionUtils.js
```diff
@@ -295,13 +295,14 @@ const validateNewVersions = async ( org_id, { channel, newVersions }, context )
   if( !newVersions || newVersions.length == 0 ) return;
 
   const { models } = context;
-  const total = await models.DeployableVersion.count( { org_id, channel_id: channel.uuid } );
+  const deployableVersionObjs = await models.DeployableVersion.find( { org_id, channel_id: channel.uuid } );
```
adamkingit, line 298: Is `find` equally efficient as `count`?

app/apollo/resolvers/channel.js
carrolp: It's not exactly an 'expected' failure -- it would typically mean something like mongo accepted the save of one record, then failed to accept the second.  It's continuable in the sense that we can't back up and delete the first record (chances are mongo is malfunctioning and it would be impossible), but this should still be reported as an error rather than a warning.

app/apollo/resolvers/channel.js
carrolp: There is a difference between the two states:
- `!version` means the caller of the API did not specify the version so it was not created
- `!version.uuid` means the caller of the API provided the version details but an error occurred when trying to create it

Keeping the messages (slightly) distinct can help explain behavior if/when either is encountered.

app/apollo/utils/versionUtils.js
carrolp: Likely not -- at the very least, `find` will use more bandwidth to return the content.  However `validateNewVersions` is infrequently called (only when creating new Versions) and the impact is therefore acceptable.

adamkingit (approved):

PR #1303, opened 2023-03-03
resolver and channel additions to provide clear response data

ethanstjohn:
For: https://github.ibm.com/alchemy-containers/satellite-config/issues/1482
Connected with: https://github.ibm.com/alchemy-containers/satellite-config/pull/1695

### DIFF
--- a/app/apollo/resolvers/channel.js
+++ b/app/apollo/resolvers/channel.js
@@ -405,6 +405,9 @@ const channelResolvers = {
           }
         } ) );
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: newChannelObj.name, uuid: newChannelObj.uuid};
+
         logger.info({ req_id, user, org_id, name }, `${queryName} returning`);
         return {
           uuid: newChannelObj.uuid,
@@ -479,6 +482,9 @@ const channelResolvers = {
           logger.error({ req_id, user, org_id, uuid, name, error }, `${queryName} failed to update the channel name in versions, continuing`);
           // Cannot fail here, the Channel has already been updated.  Continue.
         }
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name, oldName: channel.name, uuid: channel.uuid};
 
         return {
           uuid,
@@ -597,6 +603,9 @@ const channelResolvers = {
             // Cannot fail here, the Version has already been created.  Continue.
           }
         } ) );
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: newVersionObj.name, uuid: newVersionObj.uuid};
 
         logger.info({req_id, user, org_id, channel_uuid, name, type }, `${queryName} returning`);
         return {
@@ -767,6 +776,9 @@ const channelResolvers = {
         } );
         */
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: channel.name, uuid: channel.uuid};
+
         logger.info({req_id, user, org_id, uuid }, `${queryName} returning`);
         return {
           success: true,
@@ -829,6 +841,9 @@ const channelResolvers = {
 
         // Deletes the configuration channel
         await models.Channel.deleteOne({ org_id, uuid });
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: channel.name, uuid: channel.uuid};
 
         logger.info({ req_id, user, org_id, uuid }, `${queryName} returning`);
         return {
@@ -908,6 +923,9 @@ const channelResolvers = {
         // Delete the Version
         await models.DeployableVersion.deleteOne({ org_id, uuid });
         logger.info({ver_uuid: uuid, ver_name: deployableVersionObj.name}, `${queryName} version deleted`);
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: channel.name, uuid: channel.uuid};
 
         logger.info({ req_id, user, org_id, uuid }, `${queryName} returning`);
         // Return success if Version was deleted
--- a/app/apollo/resolvers/cluster.js
+++ b/app/apollo/resolvers/cluster.js
@@ -423,6 +423,9 @@ const clusterResolvers = {
         const deletedResourceYamlHist = await models.ResourceYamlHist.updateMany({ org_id, cluster_id }, {$set: { deleted: true }}, { upsert: false });
         logger.info({req_id, user, org_id, cluster_id, deletedResourceYamlHist}, 'ResourceYamlHist soft-deletion complete');
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: deletedCluster.registration, uuid: deletedCluster.cluster_id};
+
         logger.info({req_id, user, org_id, cluster_id}, `${queryName} returning`);
         return {
           deletedClusterCount: deletedCluster ? (deletedCluster.cluster_id === cluster_id ? 1 : 0) : 0,
@@ -476,6 +479,9 @@ const clusterResolvers = {
         logger.info({req_id, user, org_id, deletedResources}, 'Resources soft-deletion complete');
         const deletedResourceYamlHist = await models.ResourceYamlHist.updateMany({ org_id }, {$set: { deleted: true }}, { upsert: false });
         logger.info({req_id, user, org_id, deletedResourceYamlHist}, 'ResourceYamlHist soft-deletion complete');
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: deletedClusters.registration, uuid: deletedClusters.cluster_id};
 
         logger.info({req_id, user, org_id}, `${queryName} returning`);
         return {
@@ -565,6 +571,9 @@ const clusterResolvers = {
           });
         }
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: registration, uuid: cluster_id};
+
         logger.info({req_id, user, org_id, registration, cluster_id}, `${queryName} returning`);
         return { url, orgId: org_id, clusterId: cluster_id, orgKey: bestOrgKey( org ).key, regState: reg_state, registration };
       } catch (error) {
@@ -597,6 +606,9 @@ const clusterResolvers = {
         const updatedCluster = await models.Cluster.findOneAndUpdate(
           {org_id: org_id, cluster_id: cluster_id},
           {$set: {reg_state: CLUSTER_REG_STATES.REGISTERING}});
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {uuid: cluster_id};
 
         if (updatedCluster) {
           var { url } = await models.Organization.getRegistrationUrl(org_id, context);
--- a/app/apollo/resolvers/group.js
+++ b/app/apollo/resolvers/group.js
@@ -148,6 +148,9 @@ const groupResolvers = {
 
         pubSub.channelSubChangedFunc({org_id: org_id}, context);
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name, uuid};
+
         logger.info({ req_id, user, org_id, name }, `${queryName} returning`);
         return {
           uuid,
@@ -196,6 +199,9 @@ const groupResolvers = {
         await models.Group.deleteOne({ org_id: org_id, uuid:group.uuid });
 
         pubSub.channelSubChangedFunc({org_id: org_id}, context);
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: group.name, uuid: group.uuid};
 
         logger.info({ req_id, user, org_id, uuid }, `${queryName} returning`);
         return {
@@ -259,6 +265,9 @@ const groupResolvers = {
         await models.Group.deleteOne({ org_id: org_id, uuid: group.uuid });
 
         pubSub.channelSubChangedFunc({org_id: org_id}, context);
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: group.name, uuid: group.uuid};
 
         logger.info({ req_id, user, org_id, name }, `${queryName} returning`);
         return {
@@ -351,6 +360,10 @@ const groupResolvers = {
         */
         groupsRbacSync( groups, { resync: false }, context ).catch(function(){/*ignore*/});
 
+        // !Which of these is correct?
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: groups.name, uuid: groupUuids};
+
         logger.info({ req_id, user, org_id, groupUuids, clusterIds }, `${queryName} returning`);
         return {
           modified: res.modifiedCount
@@ -398,6 +411,10 @@ const groupResolvers = {
         );
 
         pubSub.channelSubChangedFunc({org_id}, context);
+
+        // Same as assign clusterGroups, no name. groups.name would be every name not the unassigned ones.
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {uuid: groupUuids};
 
         logger.info({ req_id, user, org_id, groupUuids, clusterIds }, `${queryName} returning`);
         return {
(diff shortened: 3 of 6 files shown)


app/apollo/resolvers/cluster.js
```diff
@@ -477,6 +480,9 @@ const clusterResolvers = {
         const deletedResourceYamlHist = await models.ResourceYamlHist.updateMany({ org_id }, {$set: { deleted: true }}, { upsert: false });
         logger.info({req_id, user, org_id, deletedResourceYamlHist}, 'ResourceYamlHist soft-deletion complete');
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: deletedClusters.registration, uuid: deletedClusters.cluster_id};
```
carrolp, line 484: The `name` and `registration` object are not the same thing -- registration _includes_ name as an attribute, but it's not the name itself.  This might be more appropriate:
```suggestion
        context.pluginContext = {name: deletedClusters.registration.name, uuid: deletedClusters.cluster_id, registration: deletedClusters.registration};
```
**However**
- note that this function is `deleteClusters` and the object `deletedClusters`, both with an `s` -- it's not just deleting a single cluster it's deleting all clusters in the org/account.  The response from the mongo `deleteMany` call is not going to be a single object with a `registration` attribute.
- subscriptions and resources and resourceYamlHist are also being deleted and information about them should also be included in the `pluginContext`.

Other updates in this PR need to be reviewed to make sure they dont have the same problems.

app/apollo/resolvers/channel.js
```diff
@@ -405,6 +405,9 @@ const channelResolvers = {
           }
         } ) );
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: newChannelObj.name, uuid: newChannelObj.uuid};
```
carrolp, line 409: pluginContext needs to include details about created version(s) and subscription(s), if created.

app/apollo/resolvers/channel.js
```diff
@@ -598,6 +604,9 @@ const channelResolvers = {
           }
         } ) );
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: newVersionObj.name, uuid: newVersionObj.uuid};
```
carrolp, line 608: pluginContext needs to include details about created subscription(s), if created

carrolp (changes requested):

app/apollo/resolvers/channel.js
```diff
@@ -460,6 +463,9 @@ const channelResolvers = {
         // Save the change
         await models.Channel.updateOne({ org_id, uuid }, { $set: { name, tags, custom, remote, updated: Date.now() } }, {});
 
+        // Save previous name for qraphQL plugins
+        const previousName = channel.name;
```
carrolp, line 467: Why save here and use later rather than just (later) doing
```
        // Allow graphQL plugins to retrieve more information
        context.pluginContext = {name, previousName: channel.name, uuid: uuid};
```
?  It doesnt look like `channel.name` would somehow _change_ between the two spots, so explicitly saving the value into a new variable here just adds complexity.

ethanstjohn: You're right, I think I was thinking about each of the other functions and how they all keep context.pluginContext in the same area in code above the return.

app/apollo/resolvers/channel.js
```diff
@@ -405,6 +405,9 @@ const channelResolvers = {
           }
         } ) );
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {newChannelObj};
```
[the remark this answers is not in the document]
ethanstjohn: updated

app/apollo/resolvers/channel.js
```diff
@@ -405,6 +405,9 @@ const channelResolvers = {
           }
         } ) );
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {name: newChannelObj.name, uuid: newChannelObj.uuid, versions: versions, subscriptions: subscriptions};
```
carrolp, line 409: instead of dumping the entire subscriptions and versions objects, should use `map()` to just include their names and uuids?  Also keep in mind that the uuids will be generated in previous code and will likely not be _in/on_ the versions or subscriptions that were passed into the function

app/apollo/resolvers/cluster.js
```diff
@@ -444,6 +444,9 @@ const clusterResolvers = {
         const deletedResourceYamlHist = await models.ResourceYamlHist.updateMany({ org_id, cluster_id }, {$set: { deleted: true }}, { upsert: false });
         logger.info({req_id, user, org_id, cluster_id, deletedResourceYamlHist}, 'ResourceYamlHist soft-deletion complete');
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {registration: deletedCluster.registration, uuid: deletedCluster.cluster_id};
```
carrolp, line 448: Why the whole registration instead of just
```suggestion
        context.pluginContext = {name: deletedCluster.registration.name, uuid: deletedCluster.cluster_id};
```

app/apollo/resolvers/cluster.js
```diff
@@ -588,6 +599,9 @@ const clusterResolvers = {
           });
         }
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {registration: registration, uuid: cluster_id};
```
carrolp, line 603: Why entire registration object instead of:
```suggestion
        context.pluginContext = {name: registration.name, uuid: cluster_id};
```
?

app/apollo/resolvers/cluster.js
```diff
@@ -629,10 +643,15 @@ const clusterResolvers = {
               url += `&args=${arg}`;
             });
           }
+          // Allow graphQL plugins to retrieve more information
+          context.pluginContext = {uuid: updatedCluster.cluster_id, url: url, registration: updatedCluster.registration};
```
carrolp, line 647:
```suggestion
          context.pluginContext = {uuid: cluster_id, name: updatedCluster.registration.name};
```
- Shouldn't need to include `url` in the pluginContext -- it's not of interest to plugins.
- should just use the `cluster_id` argument passed in
- should just include the name, not the entire `registration` object

app/apollo/resolvers/cluster.js
```diff
@@ -629,10 +643,15 @@ const clusterResolvers = {
...
 
           logger.info({ req_id, user, org_id, cluster_id }, `${queryName} returning`);
           return { url };
         } else {
+          // Allow graphQL plugins to retrieve more information
+          context.pluginContext = {uuid: updatedCluster.cluster_id, registration: updatedCluster.registration};
```
carrolp, line 653:
```suggestion
          context.pluginContext = {uuid: updatedCluster.cluster_id, name: updatedCluster.registration.name };
```

app/apollo/resolvers/group.js
```diff
@@ -351,6 +360,9 @@ const groupResolvers = {
         */
         groupsRbacSync( groups, { resync: false }, context ).catch(function(){/*ignore*/});
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {clusterNames: clusterIds, groupDetails: groupObjsToAdd};
```
carrolp, line 364: how can we get this to be
```
        context.pluginContext = { clusters: [], groups: [] };
```
Where `clusters` array members and groups array members are `{ uuid, name}` objects?  I _think_ the `groupObjsToAdd` object is like that, but having just cluster names or cluster ids is not sufficient.  And `clusterNames: clusterIds` also just confusing -- they're not _names_ :-).

app/apollo/resolvers/group.js
```diff
@@ -399,6 +411,9 @@ const groupResolvers = {
 
         pubSub.channelSubChangedFunc({org_id}, context);
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {clusterNames: clusterIds, groupUuids: groupUuids};
```
carrolp, line 415: `clusterNames: clusterIds`: names or uuids?  we want/need both for the context to be usable, right?

app/apollo/resolvers/group.js
```diff
@@ -465,6 +480,9 @@ const groupResolvers = {
         */
         groupsRbacSync( groups, { resync: false }, context ).catch(function(){/*ignore*/});
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {clusterId: clusterId, groupDetails: groupObjsToAdd};
```
carrolp, line 484: pluginContext should be like `{ name: clusterName, uuid: clusterId, groups: [{name,uuid}]`

app/apollo/resolvers/group.js
ethanstjohn: hmm, when I tested and ran the output for clusterIds, it did only give me the name I labeled it no id or uuid

app/apollo/resolvers/channel.js
```diff
@@ -405,6 +405,25 @@ const channelResolvers = {
...
+        const subscriptionFind = await models.Subscription.find({org_id});
+        const subscriptionObjs = _.map(subscriptionFind, (subscription)=>{
+          return {
+            name: subscription.name,
+            uuid: subscription.uuid,
+          };
+        });
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {channel: {name: newChannelObj.name, uuid: newChannelObj.uuid}, versions: versionObjs, subscriptions: subscriptionObjs};
```
carrolp, line 425: For consistency with other actions, the pluginContext should be
```suggestion
        context.pluginContext = { name: newChannelObj.name, uuid: newChannelObj.uuid, versions: versionObjs, subscriptions: subscriptionObjs};
```
(object should have name and uuid top level attributes for the channel, not in a sub attribute).

carrolp: NVM, I see now that the other resolvers now follow the same pattern

app/apollo/resolvers/channel.js
```diff
@@ -598,6 +620,18 @@ const channelResolvers = {
...
         } ) );
 
+        // Create output for graphQL plugins
+        const subscriptionFind = await models.Subscription.find({org_id});
+        const subscriptionObjs = _.map(subscriptionFind, (subscription)=>{
+          return {
+            name: subscription.name,
+            uuid: subscription.uuid,
+          };
+        });
```
carrolp, line 630: This finds all subscriptions in the org, not just those affected by this particular action. Instead of needing a new query, the subscriptions should just be built by the processing of the earlier Promise.all()'.

app/apollo/resolvers/channel.js
```diff
@@ -767,6 +801,25 @@ const channelResolvers = {
...
+
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {channel: {name: channel.name, uuid: channel.uuid}, versions: versionObjs, subscriptions: subscriptionObjs};
+
```
carrolp, line 822: `editChannelVersion` **only** affects a single version record, right?  So the pluginContext shouldn't need channel or subscription details..... it should just pull the name and uuid out of the `version` object retrieved earlier in the function.  No additional queries of 'all versions' or 'all subscriptions' needed.

carrolp: An argument could be made for including channel details (since a version is a 'sub object' of a Channel), but that would also not need any additional queries as the `channel` variable was already retrieved.

app/apollo/resolvers/channel.js
```diff
@@ -830,6 +883,9 @@ const channelResolvers = {
         // Deletes the configuration channel
         await models.Channel.deleteOne({ org_id, uuid });
 
+        // Allow graphQL plugins to retrieve more information
+        context.pluginContext = {channel: {name: channel.name, uuid: channel.uuid}};
```
carrolp, line 887: Deletion of a Channel **does** automatically delete its versions as well (see `deleteMany` just a couple lines up).  So this is a case where the code _should_ query all the Versions for the config being deleted and include details on those in the pluginContext as well as the channel

local-dev/api/clustersDelete.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+RAZEE_ORG_ID=${2:-${RAZEE_ORG_ID:-pOrgId}}
```
carrolp, line 5:
```suggestion
RAZEE_ORG_ID=${1:-${RAZEE_ORG_ID:-pOrgId}}
```

local-dev/api/clusterEnableRegistrationUrl.sh
```diff
@@ -0,0 +1,13 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+RAZEE_CLUSTER_UUID=${1:-${RAZEE_CLUSTER_UUID:-pClusterId}}
+RAZEE_ORG_ID=${2:-${RAZEE_ORG_ID:-pOrgId}}
+
+RAZEE_QUERY='mutation($orgId: String! $clusterId: String!) { enableRegistrationUrl( orgId: $orgId clusterId: $clusterId) { url } }'
+RAZEE_VARIABLES='{"clusterId":"'"${RAZEE_CLUSTER_UUID}"'","orgId":"'"${RAZEE_ORG_ID}"'"}'
+
+echo "" && echo "UPDATE registration url"
```
carrolp, line 11:
```suggestion
echo "" && echo "ENABLE registration url"
```

local-dev/api/configRename.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+RAZEE_CONFIG_NAME=${1:-pConfigName}
+RAZEE_CONFIG_UUID=${2:-${RAZEE_CONFIG_UUID:-pConfigId}}
+RAZEE_ORG_ID=${3:-${RAZEE_ORG_ID:-pOrgId}}
+
+RAZEE_QUERY='mutation ($orgId: String!, $uuid: String!, $name: String!) { editChannel(orgId: $orgId, uuid: $uuid, name: $name) { success } }'
+RAZEE_VARIABLES='{"name":"'"${RAZEE_CONFIG_NAME}"'","uuid":"'"${RAZEE_CONFIG_UUID}"'","orgId":"'"${RAZEE_ORG_ID}"'"}'
+
+echo "" && echo "CREATE config by name"
```
carrolp, line 16:
```suggestion
echo "" && echo "EDIT config with new name"
```

local-dev/api/groupEditCluster.sh
```diff
@@ -0,0 +1,14 @@
...
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+RAZEE_CLUSTER_UUID=${1:-${RAZEE_CLUSTER_UUID:-pClusterId}}
+RAZEE_GROUP_UUID=${2:-${RAZEE_GROUP_UUID:-pGroupId}}
+RAZEE_ORG_ID=${3:-${RAZEE_ORG_ID:-pOrgId}}
+
+RAZEE_QUERY='mutation($orgId: String!, $clusterId: String!, $groupUuids: [String!]!){ editClusterGroups( orgId: $orgId, clusterId: $clusterId, groupUuids: $groupUuids ) { modified } }'
+RAZEE_VARIABLES='{"orgId":"'"${RAZEE_ORG_ID}"'","groupUuids":["'"${RAZEE_GROUP_UUID}"'"],"clusterId":"'"${RAZEE_CLUSTER_UUID}"'"}'
+
+echo "" && echo "EDIT Group in Cluster"
```
carrolp, line 12:
```suggestion
echo "" && echo "EDIT cluster groups"
```

local-dev/api/subSetVersion.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+RAZEE_SUB_UUID=${2:-${RAZEE_SUB_UUID:-pSubUuid}}
+RAZEE_ORG_ID=${3:-${RAZEE_ORG_ID:-pOrgId}}
+RAZEE_VER_UUID=${6:-${RAZEE_VER_UUID:-pVerUuid}}
```
carrolp, line 7:
```suggestion
RAZEE_SUB_UUID=${1:-${RAZEE_SUB_UUID:-pSubUuid}}
RAZEE_ORG_ID=${2:-${RAZEE_ORG_ID:-pOrgId}}
RAZEE_VER_UUID=${3:-${RAZEE_VER_UUID:-pVerUuid}}
```

carrolp (approved):

Issue #1316, opened 2023-05-03
Bug: Log Clean-up


ethanstjohn:
Remove excess logging and potential logging of sensitive values in some error-handling scenarios.

- This was partially remediated [here](https://github.com/razee-io/Razeedash-api/pull/1314/files#diff-57d85898bb94c2e7e4aeaa615d7982b006e346c32dd69a76da2d3fc2ba811650R115-R134), reducing the amount of times `razee-org-key` shows up without '[REDACTED]'
- `razee-org-key` now only appears in SatCon LogDNA logs for `level: 40` and `status-code: 404`

PR #1317, opened 2023-05-04
redact all razee-org-keys

ethanstjohn:
fixes https://github.com/razee-io/Razeedash-api/issues/1316

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -116,6 +116,11 @@ router.use('/v1/systemSubscriptions', SystemSubscriptions);
 // This should be the last router.use -- any later additions will be restricted by the redaction.
 router.use( (err, req, res, next) => {
   if( req && req.headers && req.headers['razee-org-key']) {
+    const orgKey = req.headers['razee-org-key'];
+    if( req.url && req.url.indexOf( orgKey ) >= 0 ) {
+      const parts = req.url.split(req.query.orgKey);
+      req.url = `${parts[0]}[REDACTED]${parts[1]}`;
+    }
     req.headers['razee-org-key'] = '[REDACTED]';
   }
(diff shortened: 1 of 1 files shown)


app/routes/index.js
```diff
@@ -114,23 +114,18 @@ router.use('/v1/systemSubscriptions', SystemSubscriptions);
...
-      const orgKey = req.query.orgKey;
-      if( req.url && req.url.indexOf( orgKey ) >= 0 ) {
-        const parts = req.url.split(req.query.orgKey);
-        req.url = `${parts[0]}[REDACTED]${parts[1]}`;
-      }
-      req.query.orgKey = '[REDACTED]';
-    }
-    next(err);
+  if( req && req.headers && req.headers['razee-org-key']) {
+    req.headers['razee-org-key'] = '[REDACTED]';
```
carrolp, line 118:
```suggestion
    const orgKey = req.headers['razee-org-key'];
    if( req.url && req.url.indexOf( orgKey ) >= 0 ) {
      const parts = req.url.split(req.query.orgKey);
      req.url = `${parts[0]}[REDACTED]${parts[1]}`;
    }
    req.headers['razee-org-key'] = '[REDACTED]';
```

app/routes/index.js
```diff
@@ -114,23 +114,18 @@ router.use('/v1/systemSubscriptions', SystemSubscriptions);
 
 // Ensure sensitive information is removed before default error handler can log request details
```
carrolp, line 115:
```suggestion
// Ensure sensitive information is removed before default handler for errors, 404s, etc can log request details
// This should be the last router.use -- any later additions will be restricted by the redaction.
```

carrolp (changes requested):

PR #1318, opened 2023-05-04
tag support for subscriptions and service subscriptions

carrolp:
- Add support to create and edit subscriptions and serviceSubscriptions with tags (arrays of strings, like for Channels)
- Add support to query subscriptions and service subscriptions by matching tags, and to include tags in results
- Unit tests


app/apollo/resolvers/serviceSubscription.js
```diff
@@ -83,21 +83,24 @@ const serviceResolvers = {
 
       logger.debug({req_id, user, orgId }, `${queryName} enter`);
 
-      try {
-        await validAuth(me, orgId, ACTIONS.READ, TYPES.SERVICESUBSCRIPTION, queryName, context);
+      await validAuth(me, orgId, ACTIONS.READ, TYPES.SERVICESUBSCRIPTION, queryName, context);
```
adamkingit, line 86: Why did you move this our of the try/catch block? Does it land in the right place if validAuth throws?

adamkingit (approved):

adamkingit (approved):

PR #1341, opened 2023-10-19
Add custom metric framework and cluster API metrics

ethanstjohn:
For Add API Metrics requested: https://github.ibm.com/alchemy-containers/satellite-config/issues/1789


app/customMetricsClient.js
```diff
@@ -0,0 +1,424 @@
...
+  buckets: [0.1, 0.5, 1, 2, 5]
+});
+
+const apiCallsCounter = new Counter({
+  name: 'my_api_calls_total',
+  help: 'Total number of API calls'
+});
+
+// Gauge success and failure of signIn
+const signInGauge = new Gauge({
```
lcopi, line 34: I think these gauges should be counters so you can check rate or increase over time etc. Also nitpick on the name  it should have `*_total` in the name

app/customMetricsClient.js
```diff
@@ -0,0 +1,424 @@
...
+  name: 'sign_in_result_status',
+  help: 'Total number of signIn operations, labeled by success or failure',
+  labelNames: ['status'],
+});
+
+// --- Cluster Resolver API Metrics ---
+// Count duration of clustersByClusterId
+const clusterByClusterIdHistogram = new Histogram({
```
lcopi, line 42: I think for both of the histogram and the gauge metrics the operation can be a label on the metric so you can consolidate all of these down into a single metric definition. Where the labels indicate what the action is

app/customMetricsClient.js
```diff
@@ -0,0 +1,424 @@
...
+  name: 'cluster_by_cluster_id_duration_seconds',
+  help: 'Duration of clusterByClusterId operations in seconds',
+  buckets: [0.1, 0.5, 1, 2, 5]
+});
+
+// Gauge success and failure of clustersByClusterId
+const clusterByClusterIdGauge = new Gauge({
```
lcopi, line 49: just confirming based on the labelNames the ID is not actually part of the metric (labels or value). Which is correct

lcopi (commented):
some suggestions

app/customMetricsClient.js
ethanstjohn: Right, the ID is not part of the metric. It is just the queryName of the API action. The metric has no way of reporting the ID like this

app/customMetricsClient.js
ethanstjohn: I will take a look and see if I can consolidate, thanks!

app/customMetricsClient.js
ethanstjohn: Good idea for the name change. and I'll switch to counters

carrolp (dismissed):

carrolp (approved):

